PR #8, opened 2020-05-14
Add ldap configuration values

michaelimfeld:


charts/timed/templates/configmap-backend.yaml
```diff
@@ -20,3 +20,10 @@ data:
 {{- else }}
   DJANGO_DATABASE_HOST: {{ .Values.postgresql.service.host | quote }}
 {{- end }}
+{{- if .Values.auth.ldap.enabled }}
+  DJANGO_AUTH_LDAP_ENABLED: true
+{{- end }}
```
tongpu, line 25: `{{- end }}` should be moved to the end of the LDAP specific values.

tongpu:
Regarding the issues with pre-commit, you should ask @hairmare

charts/timed/values.yaml
```diff
@@ -158,3 +158,11 @@ postgresql:
...
     # postgresql.image.tag -- PostgreSQL image version to use
     tag: 12.2.0
+
+auth:
+  ldap:
+    enabled: false
+    # serverUri: 'ldaps://ldap.foo.com:636'
+    # bindDn: 'uid=auth_user,cn=systems,dc=foo,dc=com'
+    # userDnTemplate: 'uid=%(user)s,cn=users,dc=foo,dc=com''
+    # userAttrMap: '{"first_name": "givenName", "last_name": "sn", "email": "mail"}'
```
hairmare, line 168: The docs should be inline in the YAML like so:
```suggestion
auth:
  ldap:
    # auth.ldap.enabled -- Enable LDAP user authentication
    enabled: false
    # auth.ldap.servicerUri -- LDAP server URI
    serverUri: "ldaps://ldap.example.com:636"
    # auth.ldap.bindDn -- LDAP bind DN
    bindDn: "uid=auth_user,cn=systems,dc=example,dc=com"
    # auth.ldap.userDnTemplate -- LDAP user DN template
    userDnTemplate: "uid=%(user)s,cn=users,dc=foo,dc=com"
    # auth.ldap.userAttrMap -- LDAP user attribute mapping
    userAttrMap: '{"first_name": "givenName", "last_name": "sn", "email": "mail"}'
```
After that you can run `helm-docs` to regenerate the readme.

charts/timed/templates/deployment-backend.yaml
```diff
@@ -25,6 +25,11 @@ spec:
                 secretKeyRef:
                   name: {{ include "timed.postgresql.fullname" . }}
                   key: postgresql-password
+            - name: DJANGO_AUTH_LDAP_BIND_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "timed.fullname" . }}-backend
```
hairmare, line 31: The secret referenced here is missing. There is a secret called `{{ include "timed.fullname" . }}-django` where we could add the `ldap-apssword` key.

```suggestion
                  name: {{ include "timed.fullname" . }}-django
```

It look like that secret in `templates/pre-install-secret-backend.yaml` is a pre-install hook to ensure we only generate `DJANGO_SECRET_KEY` once, but I'm not sure that really makes sense. We could probably consider changing it to a regular secret and just have folks explicitly set `backend.service.secret` if they do not want to rotate the secret on a regular basis.

PR #16, opened 2020-05-19
Timed redmine config

michaelimfeld:
Add redmine integration configuration values according to:

https://github.com/adfinis-sygroup/timed-backend/blob/master/timed/settings.py#L244

### DIFF
--- a/charts/timed/Chart.yaml
+++ b/charts/timed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 type: application
 description: Chart for Timed application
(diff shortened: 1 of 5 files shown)


charts/timed/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 type: application
 description: Chart for Timed application
-version: 0.3.5
+version: 0.3.6
```
hairmare, line 5: For some reason it wants you to bump to 0.3.7.

PR #18, opened 2020-05-19
fix(repo): There should be a link here!

hairmare:


tongpu:
@tschitschable What is the preferred domain? https://adfinis.com or https://www.adfinis.com?

tschitschable:
> @tschitschable What is the preferred domain? https://adfinis.com or https://www.adfinis.com?

The prefered domain is https://adfinis.com - thank you!

PR #22, opened 2020-05-26
timed: Add cronjobs

### LINKED ISSUE #13: timed: CronJob containers for Jobs
**Is your feature request related to a problem? Please describe.**
timed has some `manage.py` based jobs that need to be scheduled regularly.

**Describe the solution you'd like**
There should be a bunch of CronJob ressources that schedule things like redmine sync and other jobs. At glance the following jobs that are run through manage.py exist:
* notify_supervisors_shorttime (0 8 * * 4)
* notify_reviewers_unverified --offset 5 (0 8  4 * *)
* notify_reviewers_unverified --offset 12 --message 'Bitte verbleibende Positionen verifizieren.' (0 8 11 * *)
* notify_reviewers_unverified --offset 19 --message 'Bitte sofort verifizieren!' --cc admin@example.com (0 8 18 * *)
* redmine_report (0 1 * * 1)
* notify_changed_employments (0 2 * * 1)

**Describe alternatives you've considered**
None

**Additional context**
Our current VM based timed deploy runs a bunch of jobs that need to be replicated to the helm chart.

michaelimfeld:
Add cronjobs: Fixes #13

### DIFF
--- a/charts/timed/templates/cronjob-backend.yaml
+++ b/charts/timed/templates/cronjob-backend.yaml
@@ -1,0 +1,34 @@
+{{- range $key, $val := .Values.backend.cronjobs }}
+---
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ include "timed.fullname" $ }}-{{ $key | lower }}
+  labels:
+    {{- include "timed.labels" $ | nindent 4 }}
+    app.kubernetes.io/component: {{ $key | lower | quote }}
+spec:
+  schedule: {{ $val.schedule | quote }}
+  concurrencyPolicy: Forbid
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          containers:
+            - name: {{ $.Chart.Name }}-redmine-report
+              image: "{{ $.Values.backend.image.repository }}:{{ default $.Chart.AppVersion $.Values.backend.image.tag }}"
+              imagePullPolicy: {{ $.Values.backend.image.pullPolicy }}
+              env:
+                - name: DJANGO_DATABASE_PASSWORD
+                  valueFrom:
+                    secretKeyRef:
+                      name: {{ include "timed.postgresql.fullname" $ }}
+                      key: postgresql-password
+              envFrom:
+                - secretRef:
+                    name: {{ include "timed.fullname" $ }}-django
+                - configMapRef:
+                    name: {{ include "timed.fullname" $ }}-backend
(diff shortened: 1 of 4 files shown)


charts/timed/templates/cronjob-backend.yaml
```diff
@@ -0,0 +1,33 @@
+{{- range $key, $val := .Values.backend.cronjobs }}
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ include "timed.fullname" . }}-{{ $key }}
```
hairmare, line 5: You need to use the root context inside of `range`. I think something like this works:
```suggestion
  name: {{ include "timed.fullname" $ }}-{{ $key }}
```

charts/timed/values.yaml
```diff
@@ -72,6 +72,50 @@ backend:
...
+      command:
+        - "./manage.py"
+        - "notify_supervisors_shorttime"
+    # backend.cronjobs.notifyReviewersFirst -- Notify reviewers first stage
+    notifyReviewersFirst:
+      schedule: "0 8 4 * *"
+      command:
+        - "./manage.py"
+        - "notify_reviewers_unverified"
+        - "--offset 5"
```
hairmare, line 88: You might consider splitting these
```suggestion
        - "--offset"
        # backend.cronjobs.notifyReviewersFirst[3] - some docs about offset
        - "5"
```

charts/timed/values.yaml
```diff
@@ -72,6 +72,50 @@ backend:
...
+    # backend.cronjobs.notifyReviewersSecond -- Notify reviewers second stage
+    notifyReviewersSecond:
+      schedule: "0 8 11 * *"
+      command:
+        - "./manage.py"
+        - "notify_reviewers_unverified"
+        - "--offset 5"
+        - "--message"
+        - "'Bitte verbleibende Positionen verifizieren.'"
```
hairmare, line 97: Maybe this should have some docs since it's reasonable to expect that it will be changed in non de cases.

charts/timed/templates/cronjob-backend.yaml
```diff
@@ -0,0 +1,35 @@
+{{- range $key, $val := .Values.backend.cronjobs }}
+---
+
+apiVersion: batch/v1beta1
```
hairmare, line 4: superfluous white space is superfluous
```suggestion
---
apiVersion: batch/v1beta1
```

PR #27, opened 2020-06-08
CI: Fix call to release init script

hairmare:


eyenx:
is this just a rename of a file?

PR #31, opened 2020-06-08
Abort publishing GitHub Pages if no changes where staged

hairmare:


hairmare:
> This is the last one right? :P

I hope so :facepalm:

PR #34, opened 2020-06-11
:recycle: (argoconfig) allow configuring multiple apps

hairmare:
* chart: https://github.com/hairmare/helm-charts/tree/feat/multiple-apps-in-argoconfig/charts/argoconfig

### DIFF
--- a/charts/argoconfig/README.md
+++ b/charts/argoconfig/README.md
@@ -2,7 +2,7 @@ argoconfig
 ==========
 Configure Argo CD AppProjects, Applications and more
 
-Current chart version is `0.3.1`
+Current chart version is `0.4.0`
 
 Source code can be found [here](https://github.com/adfinis-sygroup/helm-charts)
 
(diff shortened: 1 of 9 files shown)


charts/argoconfig/README.md
```diff
@@ -16,18 +16,19 @@ Source code can be found [here](https://github.com/adfinis-sygroup/helm-charts)
...
-| application.source.targetRevision | string | `"*"` | Revision of chart in source repo to install |
-| application.syncPolicy | object | deactivated | Application [Sync Policy](https://argoproj.github.io/argo-cd/user-guide/auto_sync/) |
+| apps[0].destination.namespace | string | `"default"` | Target namespace of application in Argo CD |
+| apps[0].destination.server | string | `"https://kubernetes.default.svc"` | Target cluster in Argo CD |
+| apps[0].enabled | bool | `true` | Enable creating an Application resource |
+| apps[0].name | string | `"app"` | Name of app, used as suffix to app specific resources |
+| apps[0].namespace | string | `"argocd"` | Namespace for Application resource |
+| apps[0].projectOverride | string | Fullname | Name of AppProject to install into |
+| apps[0].source.chart | string | `""` | Name of chart in source repo |
+| apps[0].source.repoURL | string | `""` | URL of sourece repo |
```
eyenx, line 29: shouldn't we also allow having an `apps[0].source.path` instead of repoURL? There might be the need for adding apps which aren't released on any charts repository, but are only available through git.

charts/argoconfig/values.yaml
```diff
@@ -28,38 +40,28 @@ appProject:
...
+  name: "app"
+  # apps[0].namespace -- Namespace for Application resource
   namespace: argocd
-  # application.projectOverride -- Name of AppProject to install into
+  # apps[0].projectOverride -- Name of AppProject to install into
   # @default -- Fullname
   projectOverride: ~
   source:
-    # application.source.repoURL -- URL of sourece repo
+    # apps[0].source.repoURL -- URL of sourece repo
```
tongpu, line 54:
```suggestion
    # apps[0].source.repoURL -- URL of source repo
```

charts/argoconfig/README.md
hairmare: This is just and example, it uses `{{- .source | toYaml | nindent 4 }}` under the hood to allow expressing anything that is a valid Argo CD Application.

eyenx: Yes I see, but it's also the README which holds all possible values, right? shouldn't we document it?

charts/argoconfig/values.yaml
```diff
@@ -28,38 +40,28 @@ appProject:
...
     repoURL: ""
-    # application.source.chart -- Name of chart in source repo
+    # apps[0].source.chart -- Name of chart in source repo
```
tongpu, line 56: I would be explicit and say `Helm chart`.

hairmare: I changed this to the exact string in the Argo CD CRD

charts/argoconfig/values.yaml
hairmare: Fixed

charts/argoconfig/README.md
hairmare: It's mostly in the README so there is an example in there. In this case the README isn't trying to be complete since it doesn't want to contain everything that argo allows.

charts/argoconfig/values.yaml
```diff
@@ -28,38 +40,34 @@ appProject:
...
-  server: https://kubernetes.default.svc
-  # cluster.config -- Cluster config (ie. contents of a KUBECONFIG)
-  config: {}
+apps:
+  example:
+    # apps.example.enabled -- Enable creating an Application resource
+    enabled: false
+    # apps.example.name -- Name of app, used as suffix to app specific
+    # resources
+    name: "false"
```
eyenx, line 49: can't we define the name of the example app "example"?

eyenx: is `$app.name` even used somewhere?

hairmare: I got rid of that when I refactored apps to a map instead of an array and should get rid of it here as well.

charts/argoconfig/README.md
hairmare: I ended up refactoring this to mostly link upstream docs and only mention things that are specific to this chart. The README is now small and sweet :)

charts/argoconfig/values.yaml
hairmare: And refactored to link upstream anyway.

charts/argoconfig/values.yaml
hairmare: And refactored to link upstream anyway.

charts/argoconfig/templates/application.yaml
```diff
@@ -1,19 +1,25 @@
...
+{{- $defaults := .Values.apps.defaults -}}
+{{- range $name, $app := .Values.apps -}}
+{{- if $app.enabled -}}
 apiVersion: argoproj.io/v1alpha1
 kind: Application
 metadata:
-  name: {{ include "argoconfig.fullname" . }}
-  namespace: {{ .Values.application.namespace }}
+  name: {{ include "argoconfig.fullname" $ }}-{{ $name }}
+  {{- with (default $defaults.namespace $app.namespace) }}
```
eyenx, line 8: noob question here, is this gonna set `$defaults.namespace` if no `$app.namespace` is defined? If yes, it's very elegant imho.

hairmare: That's what `default` is doing, yeah. It's also using `with` to ensure that no namespace is set if neither `$defaults.namespace` and `$app.namespace` is set.

hairmare:
This is somewhat further along now... #36 contains some early work on an `infra-apps` chart based on this changeset.

charts/argoconfig/templates/application.yaml
```diff
@@ -1,19 +1,31 @@
...
+    {{- merge $app.destination $defaults.destination | toYaml | nindent 4 }}
+    {{- with $app.values }}
+    helm:
+      values: |-
+        {{- . | toYaml | nindent 8 }}
+    {{- end }}
+  {{- if $app.syncPolicy }}
   syncPolicy:
-    {{- .Values.application.syncPolicy | toYaml | nindent 4 }}
+    {{- $app.syncPolicy | toYaml | nindent 4 }}
```
eyenx, line 28:
```suggestion
    {{- $app.syncPolicy | toYaml | nindent 4 }}
  {{- else }}
    {{- $defaults.syncPolicy | toYaml | nindent 4 }}
```
not sure if this is really needed tho?

hairmare: I decided against it since I think we should always configure syncPolicy on a per app base.

Adding it later is a small non-bc-breaking change anyway.

PR #36, opened 2020-06-12
feat(infra-apps) initialize infra-apps

hairmare:
* `infra-apps` chart: https://github.com/hairmare/helm-charts/tree/feat/initial-infra-apps-chart/charts/infra-apps
* [x] multiple apps in argoconfig (#34)
* [x] multiple clusters in argoconfig (#39) 
* [x] refactor CI and READMEs (#40)
* [x] refactor argoconfig to a library chart (https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r440648497, #44, #45)
* [x] multiple projects in argoconfig? (most likely default to one project per `infra-*` app)
* [x] figure namespace handling (https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r440649754, https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r440650666, https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r440651292)
  * I am going to not handle namespaces here and leave them as a separate top-level concern we can address individually  
* [x] actual infra-apps chart w/ basic apps (#36)
* [x] find/document a way to use external files/template in values (https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r440648679, https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r440649126)
  * Removed from scope of this PR, will revisit if needed later on
* [ ] complete example in charts/infra-apps/examples/cert-manager.yaml (https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r440648867)
* [ ] complete example in charts/infra-apps/examples/nginx-ingress.yaml (https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r440648993)
* [x] get rid of `infra-name` repetition in values (https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r440652817)
* [ ] create a clusterissuer for cert-manager (maybe through a dedicated chart)


charts/infra-apps/examples/argocd.yaml
```diff
@@ -0,0 +1,78 @@
...
+        server:
+          config:
+            # configure main Argo CD URL
+            url: *url
+            # add a dex.config to enable auth via an external IdP or LDAP server
+            #
+            # The example below uses and external LDAP server and needs adapting
+            # by changing the example.com endpoints as well as the searches if
+            # you are on a non AD environment.
+            dex.config: |
```
hairmare, line 25: Rewriting argoconfig to a library chart will make loading this from an external file/template possible.

charts/infra-apps/examples/argocd.yaml
```diff
@@ -0,0 +1,78 @@
...
+            ingress:
+              tls:
+                - secretName: infra-argocd-cert
+                  hosts:
+                    - *host
+          # Configure Argo CD RBAC policies
+          rbacConfig:
+            # use emails and groups from dex here
+            policy.csv: |
+              g, username@example.com, role:admin
```
hairmare, line 76: Should be an external file/template as well

charts/infra-apps/examples/cert-manager.yaml
```diff
@@ -0,0 +1,11 @@
...
+  namespaces:
+    infra-certmgr:
+      additionalLabels:
+        k8s.adfinis.com/rbac-definition: "infra-rbac"
+
+argoconfig:
+  apps:
+    infra-certmgr:
+      enabled: true
+      values: {}
```
hairmare, line 11: This example needs fleshing out!

charts/infra-apps/examples/nginx-ingress.yaml
```diff
@@ -0,0 +1,11 @@
...
+  namespaces:
+    infra-ingress:
+      additionalLabels:
+        k8s.adfinis.com/rbac-definition: "infra-rbac"
+
+argoconfig:
+  apps:
+    infra-ingress:
+      enabled: true
+      values: {}
```
hairmare, line 11: This example is missing!

charts/infra-apps/examples/prometheus.yaml
```diff
@@ -0,0 +1,94 @@
...
+              enabled: true
+              grafana.ini:
+                auth.ldap:
+                  enabled: true
+                  allow_sign_up: true
+                  config_file: /etc/grafana/ldap.toml
+              ldap:
+                enabled: true
+                config: |-
+                  verbose_logging = true
```
hairmare, line 67: Should be an external file/template.

charts/infra-apps/templates/NOTES.txt
```diff
@@ -0,0 +1,13 @@
+The following Namespaces were configured:
+{{- range $name, $data := .Values.namespaces.namespaces }}
```
hairmare, line 2: Maybe namespaces should be created by an argo event or workflow.

charts/infra-apps/templates/_helpers.tpl
```diff
@@ -0,0 +1,63 @@
+{{/* vim: set filetype=mustache: */}}
```
hairmare, line 1: This needs to be reviewed and unused parts should be removed, the move to a library argoconfigs comes first though.

charts/infra-apps/templates/namespaces.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.namespaces.enabled }}
```
hairmare, line 1: As noted, I'm not sure if this makes sense here. Maybe namespaces should get created by argo events/workflow. The event/workflow could also take care of actually setting up `*-apps` helm charts.

charts/infra-apps/values.yaml
```diff
@@ -0,0 +1,118 @@
+namespaces:
+  # namespaces.enabled -- Enable creating infra Namespaces
+  enabled: true
+  # namespaces.namespaces -- Map of Namespaces
+  # @default -- see [values.yaml](./values.yaml)
+  namespaces:
+    infra-argocd:
+      enabled: true
```
hairmare, line 8: I don't like that this enables to many workspaces and isn't coupled to the namespaces used below.

Maybe we get rid of namespaces here with argo events/workflow or this needs refactoring once we have a library chart.

charts/infra-apps/values.yaml
```diff
@@ -0,0 +1,118 @@
...
+      values:
+        nameOverride: infra-argocd
+        global:
+          image:
+            pullPolicy: Always
+        server:
+          ingress:
+            annotations:
+              kubernetes.io/ingress.class: infra-ingress
+              kubernetes.io/tls-acme: true
```
hairmare, line 62: The `infra-argocd` repetition will get dealt with once argoconfig is a library.

charts/infra-apps/examples/argocd.yaml
hairmare: I'm reconsidering this since it would probably mean having more magic yaml merging than makes sense. Most likely I'll revisit this if we come across an example where the config is large enough so this makes more sense.

charts/infra-apps/examples/argocd.yaml
hairmare: https://github.com/adfinis-sygroup/helm-charts/pull/36#discussion_r449054303

charts/infra-apps/templates/NOTES.txt
hairmare: I removed the Namespace issue from the scope of this chart for now. I believe that we should manage namespaces (and maybe also rbac rules) somewhere else so we can isolate security concerns related to namespaces from the actual app deployments.

We will also want to use whatever solution we choose to manage namespaces for more than just infra apps and we don't want to have to reconfigure infra-apps when we need to a namespace for something. Maybe namespaces should be treated as data and not as code... I need to think about this a bit more and will gladly take pointers to ideas that help us keep namespaces somewhat dynamic while still gitopsing them..

hairmare:
This still needs some serious testing but I've decided to not keep it in draft-limbo any longer.

If someone could spare some time for an initial review I'd be happy as a cake.

eyenx:
I'll find some time this week

charts/infra-apps/templates/argocd.yaml
```diff
@@ -0,0 +1,29 @@
+{{ if .Values.argocd.enabled }}
+{{ template "argoconfig.application" (list . "infra-apps.argocd") }}
+---
```
yoan-adfinis, line 3: I wouldn't jam two resources into one file. Unless it's _n_ resources built from a loop of course.

yoan-adfinis: All those files (argocd.yml, cermgr.yml, ingress.yml, monitoring.yml, and rbac.yml) are very similar.

hairmare: Thanks for the feedback!

Them looking similar is one of the reasons I put both the application and as well as the project into the same file. It makes it easy to just `cat old | sed > new` when creating a new one.

I'll look if I can find a sensible way to have an application.yaml as well as a approject.yaml with some looping.

The application part is mostly a workaround until [ApplicationSets](https://github.com/argoproj-labs/applicationset) are more mature so we can leverage those. Maybe the AppProject should be removed from here letting us manage rbac/security related stuff through a different channel, but then again I already added rbac.yaml which moves some serious security concerns into this chart.

hairmare: The only way I can see some looping happen is with the `tpl` function which looks like a `exec(generated_string)` kind of pattern if we use it to generate templated templates.

Since this will be replaced with an ApplicationSet once that is possible, I'm going to keep this like it is. I hope keeping it this way will make it easier to migrate to such an ApplicationSet down the line.

eyenx:
Hmm I would suggest renaming some of this charts and values.

`monitoring` > `prometheusOperator`
`certmgr` > `certManager`
`ingress` >  `ingressNginx`
`rbac` > `rbacManager` 

For me it is clearer like that. What if we want to have several `logging` charts, a.e: `elk` & `loki`? Or what if we want to support several `ingress` providers? 

Same goes for storage. We can't just name the storage Provider `storage`. It should be named a.e `NfsClientProvisioner` or `cephCSI` or whatever it really is.

hairmare:
> Hmm I would suggest renaming some of this charts and values.
> 
> `monitoring` > `prometheusOperator`

Should be `kubePrometheusStack` if we want to name it like the chart it uses.

> `certmgr` > `certManager`
> `ingress` > `ingressNginx`
> `rbac` > `rbacManager`

tbh I took care at using the shortest names I could think of to not run into any issues with derived names and their length. I'll spin up a local cluster to do some testing before I rename them.

> For me it is clearer like that. What if we want to have several `logging` charts, a.e: `elk` & `loki`? Or what if we want to support several `ingress` providers?
> 
> Same goes for storage. We can't just name the storage Provider `storage`. It should be named a.e `NfsClientProvisioner` or `cephCSI` or whatever it really is.

This chart deploys a default set of things that IMO belong in every cluster, for that reason both `logging` and `storage` is out of the scope of the PR for now. IMO Logging and storage are often provided by the underlying platform and not managed by us.

In cases where we can't use `nginx-ingress` we would not activate the ingress that gets deployed with this chart and replace it with an env specific chart that deploys whatever ingress solution we need.

hairmare:
> Hmm I would suggest renaming some of this charts and values.

I renamed them according to your suggestions (exept prom operator which is now `kubePrometheusStack`).

charts/infra-apps/examples/rbac-manager.yaml
```diff
@@ -0,0 +1,3 @@
+rbac:
```
eyenx, line 1: shoulldn't this be `rbacManager:`

PR #44, opened 2020-06-30
:sparkles: (common) initial common chart

hairmare:
Initial chart based on parts of incubator/common. The co authors on this
PR reflect the original authors on https://github.com/helm/charts.

Compared to the incubator version this has been stripped of some parts
that we currently do not have a use for.

* [x] add common chart
* [x] strip common chart of templates we don't currently use
* [x] reconfigure chart-testing GitHub Action to ignore library chart

I created an issue about testing library charts: https://github.com/helm/chart-testing/issues/237 
before I ignored the common chart.

There is a example of how this is used in https://github.com/adfinis-sygroup/helm-charts/pull/36


charts/common/templates/_fullname.tpl
```diff
@@ -0,0 +1,29 @@
...
+*/ -}}
+{{- define "common.fullname"}}
+  {{- $global := default (dict) .Values.global -}}
+  {{- $base := default (printf "%s-%s" .Release.Name .Chart.Name) .Values.fullnameOverride -}}
+  {{- $gpre := default "" $global.fullnamePrefix -}}
+  {{- $pre := default "" .Values.fullnamePrefix -}}
+  {{- $suf := default "" .Values.fullnameSuffix -}}
+  {{- $gsuf := default "" $global.fullnameSuffix -}}
+  {{- $name := print $gpre $pre $base $suf $gsuf -}}
+  {{- $name | lower | trunc 54 | trimSuffix "-" -}}
```
eyenx, line 28: why `54` ?

eyenx: > Output of this function is truncated at 54 characters, which leaves 9 additional characters for customized overriding. Thus you can easily extend this name in your own charts:

PR #54, opened 2020-07-21
Postgresql Ingress

michaelimfeld:


charts/timed/templates/service-postgresql.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.postgresql.ingress.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: postgresql-ingress
+  labels:
+    {{- include "timed.labels" . | nindent 4 }}
+    app: postgresql
```
hairmare, line 8: Can't this be done with a [common label](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)? (ie. one with a app.kubernetes.io prefix)

Otherwise we should use something with a `k8s.adfinis.com` prefix.

charts/timed/templates/service-postgresql.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  type: LoadBalancer
+  ports:
+    - port: 5432
+      targetPort: 5432
+      protocol: TCP
+      name: tcp-postgresql
+  selector:
+    app: postgresql
+    role: slave
```
hairmare, line 18: Same here although it looks like the upstream postgresql chart is already using legacy labels.

Maybe the `timed.selectorLabels` named template helper can also be part of this?

To account for deployments that do not use the postgresql chart as a dependency the deployment might need to have a possibility to override the labels.

charts/timed/values.yaml
```diff
@@ -221,6 +221,11 @@ postgresql:
   image:
     # postgresql.image.tag -- PostgreSQL image version to use
     tag: 12.2.0
+  ingress:
+    # postgresql.ingress.enabled -- Enable ingress
+    enabled: false
+    # postgresql.ingress.ipWhitelist -- Whitelist specific IP ranges
+    ipWhitelist: []
```
hairmare, line 228:
```suggestion
    loadBalancerSourceRanges: []
```

charts/timed/templates/service-postgresql.yaml
```diff
@@ -0,0 +1,22 @@
...
+  {{- if .Values.postgresql.ingress.ipWhitelist }}
+  loadBalancerSourceRanges: {{ .Values.postgresql.ingress.ipWhitelist | toYaml | nindent 4 }}
+  {{- end }}
```
hairmare, line 21: I'm not sure if the nindent on the same line as the value will work.
```suggestion
  {{- if .Values.postgresql.ingress.loadBalancerSourceRanges }}
  loadBalancerSourceRanges: 
    {{ .Values.postgresql.ingress.loadBalancerSourceRanges | toYaml | nindent 4 }}
  {{- end }}
```

charts/timed/templates/service-postgresql.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if .Values.postgresql.ingress.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: postgresql-ingress
+  labels:
+    {{- include "timed.labels" . | nindent 4 }}
+    {{ toYaml .Values.postgresql.ingress.labels | nindent 4 }}
```
hairmare, line 8: These don't hurt but are probably also not needed.

charts/timed/values.yaml
```diff
@@ -221,6 +221,18 @@ postgresql:
   image:
     # postgresql.image.tag -- PostgreSQL image version to use
     tag: 12.2.0
+  ingress:
+    # postgresql.ingress.enabled -- Enable ingress
+    enabled: false
+    # postgresql.ingress.loadBalancerSourceRanges -- Whitelist specific IP ranges
+    loadBalancerSourceRanges: []
+    # postgresql.ingress.labels -- Service labels
```
hairmare, line 229: I would nitpick about these missing a `@default` annotation to ensure that the defaults don't result in a huge block of JSON inlined into the docs, but then I looked at the rest of the chart which doesn't care either.

PR #55, opened 2020-07-28
Timed django admins

michaelimfeld:
Add Django admins to configmap


yoan-adfinis:
would it make sense to document the expected format?

https://github.com/adfinis-sygroup/timed-backend/blob/3eceea7cf6f09bcfe23624f0c69a12b275a15a6f/timed/settings.py#L257

PR #57, opened 2020-07-29
Timed frontend oidc config

michaelimfeld:
Depends on:
https://github.com/adfinis-sygroup/timed-frontend/commit/95a144197365d42af8bc3858aca2bdab86b25929

### DIFF
--- a/charts/timed/Chart.yaml
+++ b/charts/timed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.3.17
+version: 0.3.18
 appVersion: v1.0-rc3
 keywords:
   - timed
--- a/charts/timed/README.md
+++ b/charts/timed/README.md
@@ -2,7 +2,7 @@ timed
 =====
 Chart for Timed application
 
-Current chart version is `0.3.17`
+Current chart version is `0.3.18`
 
 
 **Homepage:** <https://github.com/adfinis-sygroup/timed-frontend>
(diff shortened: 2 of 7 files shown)


charts/timed/templates/configmap-backend.yaml
```diff
@@ -16,7 +16,7 @@ data:
   DJANGO_DATABASE_USER: {{ .Values.postgresql.postgresqlUsername | quote }}
   DJANGO_DATABASE_PORT: {{ .Values.postgresql.service.port | quote }}
   DJANGO_WORK_REPORT_PATH: {{ .Values.backend.settings.workReportPath | quote }}
-  DJANGO_OIDC_USERINFO_ENDPOINT: {{ .Values.auth.oidc.userinfoEndpoint | quote }}
+  DJANGO_OIDC_USERINFO_ENDPOINT: {{ printf "%s/userinfo" .Values.auth.oidc.url | quote }}
```
hairmare, line 19: This is specific to our using it with keycloak and we'll need to change it if we end up using a different IdP.

charts/timed/templates/configmap-backend.yaml
```diff
@@ -25,7 +25,7 @@ data:
   DJANGO_OIDC_LASTNAME_CLAIM: {{ .Values.auth.oidc.claims.lastname | quote }}
 {{- if .Values.auth.oidc.introspect.enabled }}
   DJANGO_OIDC_CHECK_INTROSPECT: "True"
-  DJANGO_OIDC_INTROSPECT_ENDPOINT: {{ .Values.auth.oidc.introspect.endpoint | quote }}
+  DJANGO_OIDC_INTROSPECT_ENDPOINT: {{ printf "%s/token/introspect" .Values.auth.oidc.url | quote }}
```
hairmare, line 28: This is specific to our using it with keycloak and we'll need to change it if we end up using a different IdP.

PR #61, opened 2020-07-31
Timed psql tls

michaelimfeld:


charts/timed/values.yaml
```diff
@@ -231,6 +231,25 @@ postgresql:
...
+    certificate:
+      # postgresql.tls.certificate.name -- Name of certificate
+      name: psql-server
+      # postgresql.tls.certificate.dnsNames -- DNS names of certificate
+      dnsNames: []
+      # postgresql.tls.certificate.issuerRef -- Issuer ref
+      issuerRef:
+        group: cert-manager.io
+        kind: ClusterIssuer
+        name: letsencrypt-prod
```
hairmare, line 252: I'd like to see issuerRef configurable so we can also support other issuers.

Also, please add a default doc snippet so helm-docs doesn't dump json to the docs.

charts/timed/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.3.20
+version: 0.3.21
```
tongpu, line 5: Because you're introducing a new feature you should bump the minor release.

hairmare: Strictly speaking SemVer only mandates such a bump after reaching the 1.x version. In this case it makes sense to bump minor anyhow ;)

charts/timed/values.yaml
```diff
@@ -221,7 +221,7 @@ postgresql:
   # postgresqlPassword: s3cr3t
   image:
     # postgresql.image.tag -- PostgreSQL image version to use
-    tag: 12.2.0
+    tag: 12.3.0
```
hairmare, line 224: Can you add a quick comment as to why we are overriding the image tag and when we can get rid of the override?

Other than that, LGTM!

michaelimfeld: Added a comment. The bitnami chart uses Postgresql 11, so we probably can't get rid off it anytime soon, or am I missing something?

hairmare: I can't seem to find the comment :/ I should tell us why we are using PostgreSQL 12 and maybe hint at the fact that we can remove the tag override once the chart has official support for 12.x

PR #67, opened 2020-08-11
Timed v1.1.0

michaelimfeld:


charts/timed/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.4.1
+version: 0.4.3
 appVersion: v1.0.0
```
hairmare, line 6: :uff: this should be updated as well

PR #68, opened 2020-08-12
feat(sentry-apps): bump Sentry Helm chart to 5.0.0

yoan-adfinis:
- 5.0.0 (Sentry 20.7.2) adds Relay to the mix.
- enables NGINX by default, which has to be disabled when using Ingress.

### DIFF
--- a/charts/sentry-apps/values.yaml
+++ b/charts/sentry-apps/values.yaml
@@ -53,6 +53,12 @@ kafka:
   # kafka.transactionStateLogMinIsr -- overridden min.insync.replicas config for the transaction topic
   transactionStateLogMinIsr: 1
 
+  # nginx -- NGINX configuration (required when `ingress.enabled=false`)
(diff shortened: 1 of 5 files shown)


charts/sentry-apps/values.yaml
```diff
@@ -53,6 +53,12 @@ kafka:
   # kafka.transactionStateLogMinIsr -- overridden min.insync.replicas config for the transaction topic
   transactionStateLogMinIsr: 1
 
+  # nginx -- NGINX configuration (required when `ingress.enabled=false`)
```
hairmare, line 56:
```suggestion
# nginx -- NGINX configuration (required when `ingress.enabled=false`)
```

yoan-adfinis: I should deactivate that auto-save formatting hook for good. Thanks

PR #77, opened 2020-09-01
:boom: (kasperleyn) deprecate chart

hairmare:
This will be the last Caasperli available for testing Helm v2 charts in isolation, thus I'm deprecating it as 1.0 (feature complete).

Once we merge this and get it through CD I'll get rid of the `charts/kasperleyn` folder as well.


tongpu (approved):
I have no clue why you're deprecating it as 1.0, but other then that LGTM.

hairmare:
> I have no clue why you're deprecating it as 1.0

To enforce that we need to bump the major version if we un-deprecate it.

But also because I really wanted kasperleyn to reach 1.0.0 ;)

PR #78, opened 2020-09-01
:bug: (sentry-apps) bump with pullable version from upstream

hairmare:
😅


blaserd (approved):
I fully agree that we should bump to this upstream patch version.

PR #81, opened 2020-09-06
:bug: (sentry-apps) expose hook config

hairmare:
We need to be able to override hook values to scale down some init
jobs. Our database isn't SaaS sized and thus the job does not need
as much resources.


charts/sentry-apps/README.md
```diff
@@ -25,6 +25,7 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
 |-----|------|---------|-------------|
 | clickhouse.replicas | int | `1` | total number of clickhouse replicas |
 | config | object | See [values.yaml](https://github.com/sentry-kubernetes/charts/tree/develop/sentry) of upstream project | extra configuration for some config maps |
+| hooks | object | See [values.yaml](https://github.com/sentry-kubernetes/charts/tree/develop/sentry) of upstream project | configure hooks |
```
blaserd, line 28: The link for "values.yaml" actually directs to the directory containing this file in the upstream project. Is there a reason not to directly link to the actual file? If not, I would suggest doing so, since I would consider it to be slightly more intuitive and easier to use. (Just like it is done on line 36.)
```suggestion
| hooks | object | See [values.yaml](https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml) of upstream project | configure hooks |
```
Note that the analogous change, if desirable, should then also be applied to line 27.

charts/sentry-apps/values.yaml
```diff
@@ -98,6 +98,10 @@ system:
   # Changing it invalidates all the current sessions.
   secretKey:
 
+# hooks -- configure hooks
+# @default -- See [values.yaml](https://github.com/sentry-kubernetes/charts/tree/develop/sentry) of upstream project
```
blaserd, line 102: Basically the same detail then the one in the README.md: If there is no good reason against it, I would link directly to the values.yaml file in the upstream project.
(And most likely also adapt line 93 accordingly.)
```suggestion
# @default -- See [values.yaml](https://github.com/sentry-kubernetes/charts/blob/develop/sentry/values.yaml) of upstream project
```

blaserd:
Beside the detail(s) pointed out before, it looks good to me.

blaserd (approved):
Now this LGTM

PR #82, opened 2020-09-08
:bug: (argoconfigs) change confusing default project

hairmare:


blaserd (approved):
This increases the compatibility and the simplicity.

LGTM

PR #89, opened 2020-10-07
chore(ci): replace deprecated add-path command

hairmare:
Stomps this warning and a similiar one about env vars:
image

* [add-path docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#adding-a-system-path)
* [set env docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable)


hairmare:
¯\_(ツ)_/¯

PR #90, opened 2020-10-07
feat(misc-apps): add sentry-kubernetes app

hairmare:
This prepares the misc-apps chart so we can use it to deploy [sentry-kubernetes](https://github.com/getsentry/sentry-kubernetes) which will ship kubernetes events to sentry.


eyenx (approved):
lgtm

PR #91, opened 2020-10-07
fix(rmd): fix path in NOTES.txt

trowik:


hairmare:
We'll replace this with a non conflicting PR

PR #100, opened 2020-10-21
Translate to english

tk0e:
Default messages should be in english. They can be overwritten in ArgoCD.


hairmare:
GitHub Actions are complaining because of a missing version update in `charts/timed/values.yaml`.

There is also some tooling (`hack/helm-docs.sh`) to update the README.md based on changes in the values + version which are also needed to make CI happy.

tk0e:
I will open an issue, so someone else can take care about it.

PR #103, opened 2020-10-22
:bug: (timed) Translate subject of reviewer notification mails

tongpu:
Follow up on #100 with correct updates to the READMEs.


tongpu:
Closed because @michaelimfeld was a tad quicker then I was with opening #102

PR #107, opened 2020-10-30
fix(rmd): add SITE_URL env variable

trowik:


charts/rmd/templates/configmap.yaml
```diff
@@ -15,6 +15,9 @@ data:
   DATABASE_USER: {{ .Values.postgresql.postgresqlUsername | quote }}
   DATABASE_PORT: {{ .Values.postgresql.service.port | quote }}
   WAIT_FOR_IT_TIMER: {{ .Values.rmd.image.waitForItTimeout | quote }}
+{{- range .Values.ingress.hosts }}
+  SITE_URL: {{ . | quote }}
```
eyenx, line 19: This will add several `SITE_URL` as `ingress.hosts` is an array. We usually only put one host inside, but we need to make sure this edge case does not happen.

I would very much appreciate more if we had a general `siteUrl` in rmds's Values.

eyenx (changes requested):

hairmare (approved):

PR #108, opened 2020-10-30
fix(rmd): add resources to values.yaml

trowik:


hairmare:
We will be configuring these values in Argo CD and keep the defaults as is.

PR #111, opened 2020-11-06
feat(security-apps): initialization

eyenx:
Initialization of security-apps holding

* [x] `dex`
* [x] `gangway`
* [x] `vault` (and it's vault-k8s implementation)
* [x] `falco`
* [x] `vault-integration` (if it's not already implemented by `vault`)


eyenx:
something fishy during rebasing...

hairmare:
Looks good so far. Do dex and gangway need to be in the same namespace or can we default to having them in their own namespaces?

eyenx:
I would suggest keeping them in the same namespaca. I do not know why keep them separate.

On 6 November 2020 18:59:55 CET, Lucas Bickel <notifications@github.com> wrote:
>Looks good so far. Do dex and gangway need to be in the same namespace
>or can we default to having them in their own namespaces?

-- 

Adfinis AG
Antonio Tauro, System Engineer, GPG KeyID: 0x0796132F0077A5F8
Güterstrasse 86 | CH-4053 Basel
Office +41 61 500 31 31 | Direct +41 61 500 31 37
www.adfinis.com

hairmare:
> I would suggest keeping them in the same namespaces. I do not know why keep them separate.

namespaces are cheap af and they will help bulding stuff according to the "least access principle" down the line.

eyenx:
> > I would suggest keeping them in the same namespaces. I do not know why keep them separate.
> 
> namespaces are cheap af and they will help bulding stuff according to the "least access principle" down the line.

Yes.

But i see this two apps going hand in hand as promOperator does with grafana etc.

I even thought about creating a login app chart, where they get created together. 

We wouldnt separate fluentd from elastic, why separate gangway from dex?

hairmare:
> We wouldnt separate fluentd from elastic, why separate gangway from dex?

If we consider namespaces to be a security boundary then we should seperate fluentd from elastic. There is no reason for elastic to be in a ns that has the same privileges as fluentd. 

With dex/gangway (in the LDAP) case only dex should be allowed to access the LDAP server but gangway should not.

> I even thought about creating a login app chart, where they get created together.

I tend to prefer the solutions where we did different charts for backend and frontend parts but have not completely made up my mind on where to set the boundary/when to spilt vs. merge.

> But i see this two apps going hand in hand as promOperator does with grafana etc.

I'd prefer to see more separation there as well but since we are using `kube-prometheus-stack` as a turnkey solution i think it's fine atm.

eyenx:
The thing about ldap access made up my mind. I will separate them next week.

Can we keep em in the same argo project tho?

hairmare:
> Can we keep em in the same argo project tho?

I would say so. Generally the argo cd project ist still a bit in limbo since we're not yet gitopsing it to the degree we should.

eyenx:
`vault-integration` is not needed as it can be done through vault' charts alone:

https://github.com/hashicorp/vault-helm/blob/master/templates/server-clusterrolebinding.yaml

charts/security-apps/templates/NOTES.txt
```diff
@@ -0,0 +1,10 @@
+The following apps are available:
+{{ if .Values.dex.enabled }}
+* dex
+{{ end }}
+{{ if .Values.gangway.enabled }}
+* gangway
+{{ end }}
+{{ if .Values.vault.enabled }}
+* vault
+{{ end }}
```
hairmare, line 10: This is still missing some of the apps.

hairmare (approved):
LGTM apart from my NOTES.txt nitpick.

PR #112, opened 2020-11-06
feat(.env): bump helm-docs version

eyenx:
bump version of helm-docs to v1.4.0


michaelimfeld (dismissed):

eyenx:
@michaelimfeld pls review again, checks should be successful now

charts/timed/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.4.10
-appVersion: v1.1.2
+version: 0.4.11
+appVersion: v1.1.3
```
michaelimfeld, line 6: Why did you bump the appVersion?

eyenx: because all app needed a version bump after changing the dependency of helm-docs (generating new README.md)

eyenx: oh my bad. AppVersion does not need to be bumped... let me fix that

eyenx: 6f63f9d fixed this

michaelimfeld: Chart version yes...

michaelimfeld (approved):

hairmare:
Should we consider removing the version of helm-docs from the end of all readmes? It feels wrong bumping all the charts if the only change to the README is the version of the tool esp. given the release cadence of helm-docs. Maybe like this:

----------------------------------------------
Autogenerated from chart metadata using [helm-docs](https://github.com/norwoodj/helm-docs)

eyenx:
yes

PR #113, opened 2020-11-06
feat(logging-apps): initialization

eyenx:
Initialization of logging-apps

* [x] loki-stack
* [x] fluentd
* [x] filebeat
* [x] ~~elastic-stack?~~


hairmare:
Maybe elastic-stack would better be in it's own `-apps` chart since it is rather complex in itself (or is there an elastic-operator-chart or something that encapsulates it's complexity).

Given the elastic stuff comes in at least 3 flavors (oss, commercial and opendistro) I'm not sure I see a pragmatic way to add it in here w/o making high level decisions that should probably not be in this chart.

eyenx:
fair enough

hairmare (approved):

PR #114, opened 2020-11-06
feat(vault-monitoring): init

eyenx:
addition of new chart vault-monitoring

* [x] write logic behind internal service (internal vault) and externalName service
* [x] ~~check if the vault-agent container can be run outside of the prometheus stack to make this a standalone chart~~
* [x] ~~write deployment teamplate if above point is correct~~
* [x] write endpoint template (if service.type: externalName is set)
* [x] write prometheusrule template
* [x] write secret template
* [x] write service template
* [x] write servicemonitor template
* [x] write testing for ci
* [x] write better, deeper documentation **for integrating with prometheus, as this is not a standalone pod**
* [x] tlsConfig.caFile in servicemonitor.yaml
* [x] name in secret.yaml


eyenx:
* [ ] `Error:  templates/: template: vault-monitoring/templates/servicemonitor.yaml:8:53: executing "vault-monitoring/templates/servicemonitor.yaml" at <4>: wrong type for value; expected string; got map[string]interface {}`

charts/vault-monitoring/values.yaml
```diff
@@ -0,0 +1,117 @@
...
+  ca_path: /etc/vault/ssl/ca.crt
+  # vault.ip -- the vault ip to connect to
+  ip: 10.1.2.3
+  # vault.scheme -- the scheme to use for connection
+  scheme: https
+  # vault.ca -- the vault CA to trust
+  ca: ""
+  auth:
+    # vault.auth.mount_path -- where the kubernetes auth is mounted on vault
+    mount_path: auth/kuberntees
```
hairmare, line 95:
```suggestion
    mount_path: auth/kubernetes
```

charts/vault-monitoring/ci/default-values.yaml
```diff
@@ -0,0 +1,87 @@
...
+  service:
+    type: externalName
+    selector: {}
+  serviceMonitor:
+    labels:
+      k8s.example.com/prometheus: kube-prometheus
+
+prometheusRules:
+  enabled: true
+  rules:
```
hairmare, line 23: It could make sense to move a part of these to values.yaml since some of them (ie. `VaultSealed`) should be created on an install w/o values.

Also, this file in `ci/` should really only contain the minima needed to pass ci.

eyenx: Yeah its in the wrong file.

eyenx: Also it would generate a lot of content in the readme tho

hairmare: > Also it would generate a lot of content in the readme tho

It's enough if the readme just has a hint at the existence of the rules and maybe a link to the file.

eyenx:
Ready for Review

eyenx:
rebased because of linting

charts/vault-monitoring/ci/default-values.yaml
```diff
@@ -0,0 +1,21 @@
...
+  ca: ""
+  auth:
+    mount_path: auth/kuberntees
+    role: metrics
+  service:
+    type: ExternalName
+    selector: {}
+  serviceMonitor:
+    labels:
+      k8s.example.com/prometheus: kube-prometheus
```
hairmare, line 17: This looks kinda like it should be in `examples/` rather than `ci/`

charts/vault-monitoring/templates/endpoint.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if eq .Values.vault.service.type "externalName" }}
```
hairmare, line 1: Maybe this should have an `else` that fired an error when type is non empty and not "externalName"?

charts/vault-monitoring/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: monitoring.coreos.com/v1
```
hairmare, line 1: Should we make this optional? (ie. with `{{ if .Values.value.serviceMonitor.create }}` or similar).

That way this could also be deployed on a cluster that uses other discovery mechanisms in prom.

charts/vault-monitoring/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,31 @@
...
+kind: ServiceMonitor
+metadata:
+  name: {{ include "vaultMonitoring.name" . }}
+  labels:
+    {{- include "vaultMonitoring.labels" . | nindent 4 }}
+{{- if .Values.vault.serviceMonitor.labels }}
+    {{- toYaml .Values.vault.serviceMonitor.labels | nindent 4}}
```
hairmare, line 8: Didn't you rename lot's of stuff from `vault` to `vaultMonitoring` at some point?
```suggestion
    {{- toYaml .Values.vaultMonitoring.serviceMonitor.labels | nindent 4}}
```

charts/vault-monitoring/templates/servicemonitor.yaml
eyenx: you're right

eyenx: * [ ] TODO

charts/vault-monitoring/templates/endpoint.yaml
eyenx: * [ ] TODO

eyenx:
* [ ] rebase

charts/vault-monitoring/templates/endpoint.yaml
eyenx: There is no endpoint needed if the service is not externalName. The endpoint is created by the label matching.

hairmare: If you try specifying an invalid `externalName` an error would be nice. It should still allow deactivating the endpoint if you don't specify one.

eyenx: I don't understand. Please suggest the edit so we can move on. Thanks.

hairmare: I was thinking something like this instead of just `{{ end }}`.
```
...
{{- else if not (empty .Values.vault.service.type) }}
{{ fail (printf "Invalid service type: %s" .Values.vault.service.type) }}
{{- end }}
```

hairmare: I was thinking something like this instead of just `{{ end }}`.
```
...
{{- else if not (empty .Values.vault.service.type) }}
{{ fail (printf "Invalid service type: %s" .Values.vault.service.type) }}
{{- end }}
```

eyenx: I don't get it. What if the `vault.service.type` is `clusterIP`? why should it fail? 

`externalName` => create Endpoint + Service 
`clusterIP` => create just Service (with selectors).

hairmare: I see, I'm looking at [the docs](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) and this doesn't feel as consistent as I though it should be. I guess it's fine but I don't really understand the following:
* Why isn't it `vault.service.type=ExternalName` instead of `vault.service.type=externalName`?
* Can't we just specify a canonical DNS name directly in the service and not manage this endpoint manually?

eyenx: >     * Why isn't it `vault.service.type=ExternalName` instead of `vault.service.type=externalName`?

```yaml
type: ExternalName
```
is correct: https://kubernetes.io/docs/concepts/services-networking/service/#externalname

>     * Can't we just specify a canonical DNS name directly in the service and not manage this endpoint manually?

Prometheus ServiceMonitor needs an endpoint to work.  A lone Service without endpoint would not work for ServiceMonitor.

charts/vault-monitoring/values.yaml
```diff
@@ -0,0 +1,120 @@
...
+  scheme: https
+  # vault.ca -- the vault CA to trust -- if set a secret is created that must be mounted in Prometheus for the ServiceMonitoring to work
+  ca: ""
+  auth:
+    # vault.auth.mount_path -- where the kubernetes auth is mounted on vault
+    mount_path: auth/kubernetes
+    # vault.auth.role -- the vault role to use for connection
+    role: metrics
+  service:
+    # vault.service.type -- which type the vault service has. For connecting to an external vault server, chose externalName
```
eyenx, line 43:
```suggestion
    # vault.service.type -- which type the vault service has. For connecting to an external vault server, chose ExternalName
```

charts/vault-monitoring/README.md
```diff
@@ -0,0 +1,60 @@
...
+| vault.auth.role | string | `"metrics"` | the vault role to use for connection |
+| vault.ca | string | `""` |  |
+| vault.ca_path | string | `"/etc/vault/ssl/ca.crt"` | the CA path to include in the configuration |
+| vault.ip | string | `"10.1.2.3"` | the vault ip to connect to |
+| vault.port | int | `443` | the vault port  to connect to |
+| vault.portName | string | `"https"` | the vault portName to use in the services |
+| vault.scheme | string | `"https"` | the scheme to use for connection |
+| vault.serverName | string | `"vault.example.com"` | the vault servername |
+| vault.service.selector | object | `{}` | definition of the  vault service selector for endpoint selection. Keep empty for using ExternalName |
+| vault.service.type | string | `"ExternalName"` | which type the vault service has. For connecting to an external vault server, chose externalName |
```
eyenx, line 40:
```suggestion
| vault.service.type | string | `"ExternalName"` | which type the vault service has. For connecting to an external vault server, chose ExternalName |
```

charts/vault-monitoring/values.yaml
hairmare:
```suggestion
    # vault.service.type -- which type the vault service has. For connecting to an external vault server, choose externalName
```

hairmare (approved):
lgtm with one small nit

charts/vault-monitoring/README.md
```diff
@@ -0,0 +1,60 @@
...
+| vault.auth.role | string | `"metrics"` | the vault role to use for connection |
+| vault.ca | string | `""` |  |
+| vault.ca_path | string | `"/etc/vault/ssl/ca.crt"` | the CA path to include in the configuration |
+| vault.ip | string | `"10.1.2.3"` | the vault ip to connect to |
+| vault.port | int | `443` | the vault port  to connect to |
+| vault.portName | string | `"https"` | the vault portName to use in the services |
+| vault.scheme | string | `"https"` | the scheme to use for connection |
+| vault.serverName | string | `"vault.example.com"` | the vault servername |
+| vault.service.selector | object | `{}` | definition of the  vault service selector for endpoint selection. Keep empty for using ExternalName |
+| vault.service.type | string | `"ExternalName"` | which type the vault service has. For connecting to an external vault server, chose ExternalName |
```
eyenx, line 40:
```suggestion
| vault.service.type | string | `"ExternalName"` | which type the vault service has. For connecting to an external vault server, choose ExternalName |
```

charts/vault-monitoring/values.yaml
```diff
@@ -0,0 +1,120 @@
...
+  scheme: https
+  # vault.ca -- the vault CA to trust -- if set a secret is created that must be mounted in Prometheus for the ServiceMonitoring to work
+  ca: ""
+  auth:
+    # vault.auth.mount_path -- where the kubernetes auth is mounted on vault
+    mount_path: auth/kubernetes
+    # vault.auth.role -- the vault role to use for connection
+    role: metrics
+  service:
+    # vault.service.type -- which type the vault service has. For connecting to an external vault server, chose ExternalName
```
eyenx, line 43:
```suggestion
    # vault.service.type -- which type the vault service has. For connecting to an external vault server, choose ExternalName
```

PR #127, opened 2020-12-03
feat(csi-secret-provider-class): intialize chart

hairmare:

PR #128, opened 2020-12-03
fix(common): generate version as per helm3 create

hairmare:
`.Release.Version` was rather invalid and `helm create` uses `.Chart.AppVersion` in a conditional which makes much more sense.


hairmare:
detected when implementing #127

eyenx (approved):

PR #130, opened 2020-12-08
feat(falcosidekick): add app to security-apps chart

hairmare:


charts/security-apps/examples/falcosidekick.yaml
```diff
@@ -0,0 +1,6 @@
+falco:
```
eyenx, line 1:
```suggestion
falcosidekick:
```

eyenx (changes requested):

eyenx (approved):
:heavy_check_mark:

PR #136, opened 2020-12-17
fix(misc-apps): add name into values

eyenx:
Had the strange issue, that fullnameOverride wasn't working as expected (got
`infra-infra-downscaler` as name).

This is my output without and with setting `name` inside `values.yaml`:

```
# with
…/charts/misc-apps helm template . --set downscaler.enabled=true --set fullnameOverride=infra | grep "name: "
  name: infra-downscaler

# without
…/charts/misc-apps helm template . --set downscaler.enabled=true --set fullnameOverride=infra | grep "name: "
  name: infra-infra-downscaler
```

### DIFF
--- a/charts/misc-apps/Chart.yaml
+++ b/charts/misc-apps/Chart.yaml
@@ -3,8 +3,8 @@ name: misc-apps
 description: Argo CD app-of-apps config for miscellaneous small tools
 type: application
 # version and appVersion are in sync in this chart!
-version: 0.2.6
-appVersion: 0.2.6
+version: 0.2.7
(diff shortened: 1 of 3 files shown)


hairmare:
I'm trying to remember why this wasn't already set as i vaguely remember there being a reason.

eyenx:
it's in `infra-apps` and in `logging-apps` and `security-apps` so why not here?

hairmare:
You're right, there was a reason for setting it and we forgot here.

hairmare (approved):

PR #138, opened 2020-12-17
feat(tracing-apps): add jaeger-operator

hairmare:


eyenx:
Not sure if it's the right module to package this in `logging-apps` 

Why not `tracing-apps`

eyenx:
Thoughts @tongpu

charts/logging-apps/Chart.yaml
```diff
@@ -3,8 +3,8 @@ name: logging-apps
 description: Argo CD app-of-apps config for logging applications
 type: application
 # version and appVersion are in sync in this chart!
-version: 0.0.2
-appVersion: 0.0.1
+version: 0.1.0
```
eyenx, line 6: @tongpu and I were discussing this last week. 

Do we need to do a minor release bump when ADDING a new app? Or would a patchversion bump suffice? What are your thoughts @hairmare ?

hairmare: As per semver, bumping the minor version indicates that a new feature was added. A patch bump would indicate that we are fixing an issue.

eyenx:
no example needed in `charts/logging-apps/examples/jaegerOperator.yaml` ?

charts/logging-apps/Chart.yaml
eyenx: Okay this makes totally sense.

hairmare:
> no example needed in `charts/logging-apps/examples/jaegerOperator.yaml` ?

none available since I haven't really played with it to figure out how a complete example would look.

hairmare:
> Not sure if it's the right module to package this in `logging-apps`
> 
> Why not `tracing-apps`

Maybe we should rename `logging-apps` to `instrumentation-apps`... I'm a bit wary of us creating a new `-apps` chart for each new thing but am fine with either solution.

eyenx:
> > Not sure if it's the right module to package this in `logging-apps`
> > Why not `tracing-apps`
> 
> Maybe we should rename `logging-apps` to `instrumentation-apps`... I'm a bit wary of us creating a new `-apps` chart for each new thing but am fine with either solution.

`observability-apps` ?

hairmare:
> > > Not sure if it's the right module to package this in `logging-apps`
> > > Why not `tracing-apps`
> > 
> > 
> > Maybe we should rename `logging-apps` to `instrumentation-apps`... I'm a bit wary of us creating a new `-apps` chart for each new thing but am fine with either solution.
> 
> `observability-apps` ?

`o11y-apps` :blush:

hairmare:
If we rename it to observability we would also need to consider putting monitoring stuff into the same `-apps`. I'll most likely end up doing a new `tracing-apps` but I'm trying to think of some others things that would fit into the same chart as well before I do so.

tongpu:
> > > > Not sure if it's the right module to package this in `logging-apps`
> > > > Why not `tracing-apps`
> > > 
> > > 
> > > Maybe we should rename `logging-apps` to `instrumentation-apps`... I'm a bit wary of us creating a new `-apps` chart for each new thing but am fine with either solution.
> > 
> > 
> > `observability-apps` ?
> 
> `o11y-apps`

I am Lukas Grossar and I approve this message.

charts/logging-apps/values.yaml
```diff
@@ -41,7 +41,7 @@ filebeat:
 # fluentd -- [fluentd](https://github.com/fluent/fluentd) ([example](./examples/fluentd.yaml))
 # @default -- -
 fluentd:
-  # fluentd.enabled -- Enable loki
+  # fluentd.enabled -- Enable fluentd
```
tongpu, line 44: I guess fixing this typo was introduced by `helm-docs`?

tongpu (approved):
LGTM

charts/logging-apps/values.yaml
hairmare: It was probably copy-pasta from when we added fluentd.

PR #139, opened 2020-12-17
fix(release.yaml): adding our repo fixes the issue with ci

eyenx:
After adding the repo first with helm, the chart can be packaged.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,8 +26,9 @@ jobs:
         with:
           version: v3.4.0
 
-      - name: Add charts.adfinis.com repo
-        run: helm repo add adfinis https://charts.adfinis.com
+      - name: Add dependency chart repos
+        run: |
+          helm repo add adfinis https://charts.adfinis.com
 
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.1.0
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,6 +26,9 @@ jobs:
         with:
           version: v3.4.0
 
+      - name: Add charts.adfinis.com repo
+        run: helm repo add adfinis https://charts.adfinis.com
+
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.1.0
         env:


hairmare:
The demo only does stuff like this for actual deps: https://github.com/helm/charts-repo-actions-demo/blob/1a9640d9983d983404862daee9563fd871db3874/.github/workflows/release.yaml#L45-L47

I'd need to test it but it might be that specifying it as option might also make sense:
```yaml
      - name: Run chart-releaser
        uses: helm/chart-releaser-action@v1.1.0
        with:
          charts_repo_url: https://charts.adfinis.com
        env:
          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
```

eyenx:
> The demo only does stuff like this for actual deps: https://github.com/helm/charts-repo-actions-demo/blob/1a9640d9983d983404862daee9563fd871db3874/.github/workflows/release.yaml#L45-L47
> 
> I'd need to test it but it might be that specifying it as option might also make sense:
> 
> ```yaml
>       - name: Run chart-releaser
>         uses: helm/chart-releaser-action@v1.1.0
>         with:
>           charts_repo_url: https://charts.adfinis.com
>         env:
>           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
> ```

we have deps in our charts.

```
dependencies:
  - name: argoconfig
    version: 0.7.0
    repository: https://charts.adfinis.com
```

I already tested it locally, without adding the repo I get the same error as in ci. Adding it solves the issue:

```
/repo # cr package charts/misc-apps/
Error: no repository definition for https://charts.adfinis.com
Usage:
  cr package [CHART_PATH] [...] [flags]

Flags:
  -h, --help                     help for package
      --key string               Name of the key to use when signing
      --keyring string           Location of a public keyring (default "/root/.gnupg/pubring.gpg")
  -p, --package-path string      Path to directory with chart packages (default ".cr-release-packages")
      --passphrase-file string   Location of a file which contains the passphrase for the signing key. Use '-' in order to read from stdin
      --sign                     Use a PGP private key to sign this package

Global Flags:
      --config string   Config file (default is $HOME/.cr.yaml)

/repo # ls
CODEOWNERS                      Makefile                        docs                            linux-amd64
CONTRIBUTING.md                 README.md                       hack
LICENSE                         charts                          helm-canary-linux-amd64.tar.gz
/repo # linux-amd64/helm repo add adfinis https://charts.adfinis.com
"adfinis" has been added to your repositories
/repo # cr package charts/misc-apps/
Successfully packaged chart in /repo/charts/misc-apps and saved it to: .cr-release-packages/misc-apps-0.2.7.tgz
```

hairmare (approved):
:shipit:

PR #141, opened 2020-12-18
fix(values.yaml): remove trailing slash from repoURL

eyenx:
Error in argocd if project sourceRepos do not have the trailing slash.


tongpu (approved):
LGTM

PR #149, opened 2020-12-31
fix(templates/falcosidekick.yaml): Template values

eyenx:
Fixes falcosidekick template values, not generating the app at all.

This was a TYPO in the initial PR: https://github.com/adfinis-sygroup/helm-charts/pull/130


fujexo (approved):
LGTM

PR #150, opened 2020-12-31
feat(falco-exporter): addition of new app

eyenx:
Addition of falco-exporter. Can be used to export falco metrics to prometheus, and adding a grafana dashboard

### DIFF
--- a/charts/security-apps/templates/falco-exporter.yaml
+++ b/charts/security-apps/templates/falco-exporter.yaml
@@ -1,0 +1,33 @@
+{{ if .Values.falcoExporter.enabled }}
+{{ template "argoconfig.application" (list . "security-apps.falcoExporter") }}
+{{ end }}
+
+{{- define "security-apps.falcoExporter" -}}{{- $app := unset .Values.falcoExporter "enabled" -}}{{- $name := default $app.destination.namespace $app.name -}}
+metadata:
+  name: {{ template "common.fullname" . }}-{{ $name }}
+spec:
+  {{- if $app.project }}
+  project: {{ $app.project | quote }}
+  {{- end }}
+  source:
+    repoURL: {{ $app.repoURL | quote }}
+    chart: {{ $app.chart | quote }}
+    targetRevision: {{ $app.targetRevision | quote }}
+    helm:
+      releaseName: {{ $name | quote }}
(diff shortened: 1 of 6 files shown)


charts/security-apps/templates/falco-exporter.yaml
```diff
@@ -0,0 +1,33 @@
...
+    helm:
+      releaseName: {{ $name | quote }}
+      values: |-
+        nameOverride: {{ $name | quote }}
+        {{- $app.values | toYaml | nindent 8 }}
+  {{- if $app.destination }}
+  destination:
+    {{ $app.destination | toYaml | nindent 4 }}
+  {{- end }}
+  {{- if $app.syncPolicy }}
```
tongpu, line 25: Where are these options coming from? I don't see them in `values.yaml`?

eyenx: From the `argoconfig` library chart. 

https://github.com/adfinis-sygroup/helm-charts/blob/master/charts/argoconfig/templates/_application.yaml

tongpu: Should we make them explicitly visible by putting them in `values.yaml` of our `*-apps` charts? Otherwise people might not realize that they can be configured?

eyenx: I don't think so. I think @hairmare has the same opinion on that, has he build this in the first place.

When using dependencies you aren't really obliged to put the default values of a dep inside your own `values.yaml` kube-prometheus-stack does not do so too when using grafana subchart a.e.

eyenx:
rebased because of linting

charts/security-apps/templates/falco-exporter.yaml
tongpu: Fine for me, but then we should extend the documentation for the `argoconfig` chart to explain how to use it.

tongpu (approved):
LGTM

charts/security-apps/templates/falco-exporter.yaml
eyenx: Exactly. We might need a sync with @hairmare next week perhaps regarding this.

PR #152, opened 2021-01-07
chore(ghcr.io): move image repo

eyenx:
As `adfinissygroup/back8sup` moved to `ghcr.io/adfinis-sygroup/back8sup` this makes sure the charts uses the correct image

### DIFF
--- a/charts/back8sup/Chart.yaml
+++ b/charts/back8sup/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: back8sup
 description: Deploy back8sup to a Kubernetes Cluster
 type: application
-version: 0.2.1
-appVersion: v0.6.1
+version: 0.2.2
+appVersion: v0.7.1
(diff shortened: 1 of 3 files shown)


charts/back8sup/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: back8sup
 description: Deploy back8sup to a Kubernetes Cluster
 type: application
-version: 0.2.1
```
tongpu, line 5: I would bump to 0.3.0, because we increase by minor version in the `appVersion` and switch the docker repository, which could also cause issues for users.

tongpu (commented):
Just my opinion regarding chart version bump, feel free to ignore it.

charts/back8sup/Chart.yaml
eyenx: you're right.

tongpu (approved):
LGTM

PR #153, opened 2021-01-08
[gh-actions] feat(.env): Introduce KIND_VERSION

eyenx:
What do you guys think? Shouldn't we be testing our Charts on a Version of K8S where we are also using the charts, or should we keep it on "latest"?


tongpu:
> What do you guys think? Shouldn't we be testing our Charts on a Version of K8S where we are also using the charts, or should we keep it on "latest"?

Given that the Kubernetes API is not going through massive changes recently and any production clusters we're going to use should only be a few versions back I don't think that we really need to use a fixed version (because our clusters are also not all going to be on the same version).

hairmare:
I think we should use the default bundled kind from the helm GitHub Action. This way we can defer picking a version to upstream. We could also put all of this into a matrix and support multiple versions but that seems like it might make CI slow..

eyenx:
Closing this then.

PR #155, opened 2021-01-14
fix(chart-releaser): add dependency repo

eyenx:
Adds the https://raw.githubusercontent.com/kubernetes-sigs/secrets-store-csi-driver/master/charts repo to chart-releaser.

Can we put these inside a config file?


eyenx:
see https://github.com/adfinis-sygroup/helm-charts/runs/1703174070?check_suite_focus=true#step:7:11

hairmare (approved):

PR #157, opened 2021-01-15
fix(caasperli): templates / configmapList, add name for ci

eyenx:
I'm not sure why this is happening. But adding a `name:` to the configMapList, solves the issue.

I removed everything beneith (keeping a very simple configmap with hard coded names) and it was still failing. This is when I realized the configMapList is in the eyes of helm also just a Object, but has no name defined (does not need it to work!).

We could migrate away from using `ConfigMapList` and just use good old `---\nobject...\n---\nobject`.

I'll let the decision to you @hairmare

### DIFF
--- a/charts/caasperli/Chart.yaml
+++ b/charts/caasperli/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: caasperli
 description: Deploy Caasperli to a Kubernetes Cluster
 type: application
-version: 0.8.5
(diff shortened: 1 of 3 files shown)


eyenx:
Haha, helm lint wants a `name` but install fails because there is a `name` damn... gonna go for the 

```
----
object
----
```

approach then...

hairmare (approved):

PR #165, opened 2021-01-22
fix(quickfix/vault-monitoring): labeling issue on template of serviceMonitor

eyenx:
This was a wrong doing in our initial Chart, never tested, as default values is "no labels"


tongpu (approved):
LGTM

tongpu (approved):
LGTM

PR #171, opened 2021-02-10
Feature/improving barman chart pv handling

sinicis:
updated barman chart:
* replaced statefulset with a deployment
* pvc creation covered by helm and not by statefulset-controller

further notes:
* online resizing of pvc is still not possible on azure (https://github.com/kubernetes-sigs/azuredisk-csi-driver/issues/273)
* argo deployment in aks int test pending

### DIFF
--- a/charts/barman/Chart.yaml
+++ b/charts/barman/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: barman
 type: application
 description: Chart for Barman PostgreSQL Backup and Recovery Manager
-version: 0.0.13
+version: 0.0.14
 appVersion: "2.10"
 keywords:
   - barman
--- a/charts/barman/Chart.yaml
+++ b/charts/barman/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: barman
 type: application
 description: Chart for Barman PostgreSQL Backup and Recovery Manager
-version: 0.0.14
+version: 0.1.0
 appVersion: "2.10"
 keywords:
   - barman
--- a/charts/barman/templates/NOTES.txt
+++ b/charts/barman/templates/NOTES.txt
@@ -6,9 +6,13 @@ See: http://docs.pgbarman.org/release/2.10/#setup-of-a-new-server-in-barman
 
 Initial backup:
 
+  * Get the pod name and export it
+
+    export BARMAN_POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "barman.name" . }},app.kubernetes.io/instance={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
+
   * Obtain a bash shell on the running Barman pod.
 
-    kubectl --namespace {{ .Release.Namespace }} exec -it {{ template "barman.fullname" . }}-0 /bin/bash
+    kubectl --namespace {{ .Release.Namespace }} exec -it ${BARMAN_POD_NAME} /bin/bash
 
   * Switch to the Barman user.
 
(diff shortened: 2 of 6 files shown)


tongpu:
Please make sure to also run the `pre-commit` commands described at the end of the development workflow: https://github.com/adfinis-sygroup/helm-charts/blob/master/docs/development.md#workflow

charts/barman/templates/pvc.yaml
```diff
@@ -0,0 +1,37 @@
+{{- if or .Values.persistence.data.enabled .Values.persistence.recover.enabled }}
```
tongpu, line 1: Did you test if this `or` is required? What would happen if both are `false` and the template would solely consist of:
```yaml
---
```

charts/barman/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: barman
 type: application
 description: Chart for Barman PostgreSQL Backup and Recovery Manager
-version: 0.0.13
+version: 0.0.14
```
tongpu: Because this is a breaking change I would increase the minor version, even if we're still in the 0.x release cycle.

```suggestion
version: 0.1.0
```

tongpu (changes requested):
Look fine for the most part. Please have a look at the recommendations and make sure to run `pre-commit run -a`.

charts/barman/templates/pvc.yaml
sinicis: Nothing happens, it can handle it. The yaml-file isn't processed when both are `false` even without the `if or` statement. So I've removed it

charts/barman/templates/NOTES.txt
```diff
@@ -6,9 +6,13 @@ See: http://docs.pgbarman.org/release/2.10/#setup-of-a-new-server-in-barman
 
 Initial backup:
 
+  * Get the pod name and export it
+
+    export BARMAN_POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "barman.name" . }},app.kubernetes.io/instance={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
```
eyenx, line 11:
```suggestion
    export BARMAN_POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "barman.name" . }},app.kubernetes.io/instance={{ .Release.Name }}" -o name)
```

eyenx (dismissed):
looks good.

~~Is the `SVC/barman-exporter` not needed anymore? Or was it ever needed?~~

I just saw it's handled by the `prometheus` subfolder.

charts/barman/templates/NOTES.txt
sinicis: The default `NOTES.txt` created with a `helm create`  includes `-o jsonpath="{.items[0].metadata.name}")`. Does it really make sense to go our own way here?

sinicis:
@tongpu @eyenx What's your opinion about default resource reqeusts and limits within `values.yaml`? I'd like to set them.

charts/barman/templates/NOTES.txt
eyenx: nope. You are right.

eyenx:
> @tongpu @eyenx What's your opinion about default resource requests and limits within `values.yaml`? I'd like to set them.

Depends what barman needs for working properly. Do we have any idea?

tongpu:
> @tongpu @eyenx What's your opinion about default resource requests and limits within `values.yaml`? I'd like to set them.

Helm Charts best practices is to not configure any by default, especially if it would make it impossible to run it in small environments. If you can extract a sane default from our production environment then do it.

sinicis:
I've seen this morning in our Grafana that the max usage within the last 24 hours was 0.06 CPU and about 200MiB RAM. So my suggestions would be:
- `resources.requests.cpu` = 0.05
- `resources.limits.cpu` = 0.1
- `resources.requests.ram` = 128MiB or 64MiB
- `resources.limits.ram` = 256MiB

eyenx:
never saw a request that little `0.05` cpu.

other than that ok for me.

tongpu:
I just had another look at Grafana and would suggest this as a sane default:
```yaml
resources:
  requests:
    cpu: 100m
    ram: 128MiB
  limits:
    ram: 256MiB
```

I would not put a limit on CPU usage, because the backup service might use some CPU for compression for a short time and putting a limit on that would only make it slower.

eyenx (approved):

PR #174, opened 2021-02-17
fix(oss-challenge): add persistent volume claim

trowik:
the volume is need for uploading images for the oss-challenge

### DIFF
--- a/charts/osschallenge/README.md
+++ b/charts/osschallenge/README.md
@@ -1,6 +1,6 @@
 # osschallenge
 
-![Version: 0.1.1](https://img.shields.io/badge/Version-0.1.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: edge](https://img.shields.io/badge/AppVersion-edge-informational?style=flat-square)
+![Version: 0.1.2](https://img.shields.io/badge/Version-0.1.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: edge](https://img.shields.io/badge/AppVersion-edge-informational?style=flat-square)
 
 Chart for OSS-Challenge application
 
@@ -56,6 +56,11 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
 | osschallenge.settings.emailPort | int | `587` | Port for the SMTP server |
 | osschallenge.settings.emailUser | string | `"example@example.com"` | Username for the SMTP server |
 | osschallenge.settings.siteUrl | string | `"https://oss-challenge.ch"` | URL for OSS-Challenge |
+| persistence.accessModes | list | `["ReadWriteOnce"]` | specifes the access mode for the PV |
+| persistence.enabled | bool | `true` | specifies if persistence is enabled or not |
+| persistence.mountPath | string | `"/data"` | specifies where to mount the PV |
+| persistence.size | string | `"10Gi"` | specifies which size the PVC should request |
+| persistence.storageClassName | string | `""` | specifies which storageClassName should be used |
 | postgresql.enabled | bool | `true` | Enable PostgreSQL for persistence |
 | postgresql.postgresqlDatabase | string | `"osschallenge"` | PostgreSQL database name |
 | postgresql.postgresqlUsername | string | `"osschallenge"` | PostgreSQL user name |
--- a/charts/osschallenge/ci/default-values.yaml
+++ b/charts/osschallenge/ci/default-values.yaml
@@ -5,3 +5,6 @@ postgresql:
 osschallenge:
   settings:
     emailPassword: s3cr3t
+persistence:
+  storageClassName: "default"
+  enabled: false
--- a/charts/osschallenge/templates/pvc.yaml
+++ b/charts/osschallenge/templates/pvc.yaml
@@ -1,0 +1,17 @@
+{{- if .Values.persistence.enabled -}}
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ include "osschallenge.fullname" . }}
+  labels:
+    {{- include "osschallenge.labels" . | nindent 4 }}
+spec:
+  {{- if .Values.persistence.storageClassName }}
+  storageClassName: {{ .Values.persistence.storageClassName }}
+  {{- end }}
+  accessModes:
+  {{ .Values.persistence.accessModes | toYaml | nindent 2 }}
+  resources:
+    requests:
+      storage: {{ .Values.persistence.size }}
+{{- end -}}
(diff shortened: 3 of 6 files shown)


charts/osschallenge/README.md
```diff
@@ -56,6 +56,8 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
 | osschallenge.settings.emailPort | int | `587` | Port for the SMTP server |
 | osschallenge.settings.emailUser | string | `"example@example.com"` | Username for the SMTP server |
 | osschallenge.settings.siteUrl | string | `"https://oss-challenge.ch"` | URL for OSS-Challenge |
+| persistence.mountPath | string | `"/data"` | specifies where to mount the PV |
+| persistence.size | string | `"10Gi"` | specifies which size the PVC should request |
```
eyenx, line 60: I'm missing the documentation of `persistence.storageClassName` Have you used `helm-docs` for this?

charts/osschallenge/templates/pvc.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: {{ include "osschallenge.fullname" . }}
+  labels:
+    {{- include "osschallenge.labels" . | nindent 4 }}
+spec:
+  {{- if .Values.persistence.storageClassName }}
+  storageClassName: {{ .Values.persistence.storageClassName }}
+  {{- end }}
+  accessModes:
+  - ReadWriteOnce
```
eyenx, line 12: the `replicaCount` would be obsolete at this point.

eyenx (changes requested):

charts/osschallenge/README.md
trowik: I used the pre-commit, which does use helm-docs

charts/osschallenge/templates/pvc.yaml
trowik: So you mean, I should set replicaCount to 0 then? And why would it be obsolete?

charts/osschallenge/README.md
eyenx: can you recheck? I'm missing this: https://github.com/adfinis-sygroup/helm-charts/pull/174/files#diff-e7cb1aadd92c6e5a2393f8dddc7ceb280e160219d531b142cb850eaff7bfd7b8R113

charts/osschallenge/templates/pvc.yaml
eyenx: Not to 0, but having `ReadWriteOnce` hard coded here, would not make it possible to have more than one replica running on different nodes. And if replica aren't running on different nodes, there's really no need for it if the use case would be HA.

eyenx: > The access modes are:
    ReadWriteOnce -- the volume can be mounted as read-write by a single node
    ReadOnlyMany -- the volume can be mounted read-only by many nodes
    ReadWriteMany -- the volume can be mounted as read-write by many nodes

eyenx: Of course the storageProvider needs to support `RWX`. 

https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

charts/osschallenge/README.md
hairmare: it's commented so it wont get doxed by helm-docs

charts/osschallenge/templates/pvc.yaml
hairmare: something like this:
```suggestion
  {{ .Values.persistance.accessModes | toYaml | nindent 2 }}
```

with values:
```yaml
persistance:
  accessModes:
  - ReadWriteOnce
```

eyenx: persistence !

hairmare: oops

charts/osschallenge/README.md
trowik: Would it be better to uncomment it, set it to nil and make a default-values.yml for the CI?

eyenx: Yes.

`storageClassName: ""`

trowik: Ok, thanks. Did it like that.

trowik:
I'm not sure why the CI fails. Is it because of line 12?

charts/osschallenge/values.yaml
```diff
@@ -104,3 +104,13 @@ postgresql:
...
   # postgresql.postgresqlPassword -- Set a password for PostgreSQL
   # postgresqlPassword: ~
+
+persistence:
+  # persistence.mountPath -- specifies where to mount the PV
+  mountPath: "/data"
+  # persistence.size -- specifies which size the PVC should request
+  size: "10Gi"
+  # persistence.accessMode -- specifes the access mode for the PV
+  accessMode: ReadWriteOnce
```
hairmare, line 114:
```suggestion
  accessMode:
  - ReadWriteOnce
```

charts/osschallenge/templates/pvc.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: {{ include "osschallenge.fullname" . }}
+  labels:
+    {{- include "osschallenge.labels" . | nindent 4 }}
+spec:
+  {{- if .Values.persistence.storageClassName }}
+  storageClassName: {{ .Values.persistence.storageClassName }}
+  {{- end }}
+  accessModes:
+  {{ .Values.persistence.accessMode | toYaml | nindent 2 }}
```
hairmare, line 12:
```suggestion
  {{ .Values.persistence.accessModes | toYaml | nindent 2 }}
```

hairmare (commented):
The CI environment does not have a storage engine that can fullfill a PVC so wou'll need to disable this part in `ci/` like you did with the postgresql volume in the rmd chart.

charts/osschallenge/ci/default-values.yaml
```diff
@@ -5,3 +5,6 @@ postgresql:
 osschallenge:
   settings:
     emailPassword: s3cr3t
+persistence:
+  storageClassName: "default"
+  enabled: false
```
trowik, line 10: This should disable it, right?

hairmare: You need to implement the flag in the `pvc.yaml` template as well:
```
{{- if .Values.persistence.enabled }}
apiVersion: v1
# ...
{{- end }}
```

trowik: ah ok, I see

hairmare: You need to wrap the actual volume stuff in the deployment in a similar `if` block. I think the postgresql chart has an example where an empty volume is used if persistence is disabled.

hairmare (approved):

eyenx (approved):

PR #175, opened 2021-02-18
feat(dummyDeployment): addition

eyenx:
Adds a dummyDeployment Option to be able to make the vault-secret-store provider create the secret if extraVolumes can't be used inside the Chart/App.

### DIFF
--- a/charts/csi-secret-provider-class/README.md
+++ b/charts/csi-secret-provider-class/README.md
@@ -25,7 +25,7 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
 |-----|------|---------|-------------|
 | dummyDeployment.affinity | object | {} | sets affinity rules on the  deployment |
 | dummyDeployment.enabled | bool | false | enable the dummy deployment |
-| dummyDeployment.image.pullPolicy | string | `"ifNotPresent"` | When to pull the container image @defualt -- ifNotPresent |
+| dummyDeployment.image.pullPolicy | string | ifNotPresent | When to pull the container image |
 | dummyDeployment.image.repository | string | k8s.gcr.io/pause | Container image to deploy |
 | dummyDeployment.image.tag | string | 3.4.1 | sets the image tag to use |
 | dummyDeployment.nodeSelector | object | {} | sets a nodeSelector on the  deployment |
--- a/charts/csi-secret-provider-class/README.md
+++ b/charts/csi-secret-provider-class/README.md
@@ -1,6 +1,6 @@
 # csi-secret-provider-class
 
-![Version: 0.1.1](https://img.shields.io/badge/Version-0.1.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.1.0](https://img.shields.io/badge/AppVersion-0.1.0-informational?style=flat-square)
+![Version: 0.2.0](https://img.shields.io/badge/Version-0.2.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.2.0](https://img.shields.io/badge/AppVersion-0.2.0-informational?style=flat-square)
 
 A Helm chart to create a SecretProviderClass resource
 
@@ -17,12 +17,21 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
 
 | Repository | Name | Version |
 |------------|------|---------|
-| https://charts.adfinis.com | common | 0.0.5 |
+| https://charts.adfinis.com | common | 0.0.6 |
 
 ## Values
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| dummyDeployment.affinity | object | {} | sets affinity rules on the  deployment |
+| dummyDeployment.enabled | bool | false | enable the dummy deployment |
+| dummyDeployment.image.pullPolicy | string | `"ifNotPresent"` | When to pull the container image @defualt -- ifNotPresent |
+| dummyDeployment.image.repository | string | k8s.gcr.io/pause | Container image to deploy |
+| dummyDeployment.image.tag | string | 3.4.1 | sets the image tag to use |
+| dummyDeployment.nodeSelector | object | {} | sets a nodeSelector on the  deployment |
+| dummyDeployment.podAnnotations | object | {} | sets Pod annotations on the deployment |
+| dummyDeployment.resources | object | {} | sets resources like limits and requests on the  deployment |
+| dummyDeployment.tolerations | list | [] | sets tolerations on the  deployment |
 | parameters | object | `{}` | parameters for provider |
 | parametersTpl | string | raw values from `parameters` | template to generate parameters |
 | provider | string | `nil` | specify secret provider |
--- a/charts/csi-secret-provider-class/README.md
+++ b/charts/csi-secret-provider-class/README.md
@@ -25,7 +25,7 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
 |-----|------|---------|-------------|
 | dummyDeployment.affinity | object | {} | sets affinity rules on the  deployment |
 | dummyDeployment.enabled | bool | false | enable the dummy deployment |
-| dummyDeployment.image.pullPolicy | string | ifNotPresent | When to pull the container image |
+| dummyDeployment.image.pullPolicy | string | IfNotPresent | When to pull the container image |
 | dummyDeployment.image.repository | string | k8s.gcr.io/pause | Container image to deploy |
 | dummyDeployment.image.tag | string | 3.4.1 | sets the image tag to use |
 | dummyDeployment.nodeSelector | object | {} | sets a nodeSelector on the  deployment |
--- a/charts/csi-secret-provider-class/README.md
+++ b/charts/csi-secret-provider-class/README.md
@@ -23,15 +23,15 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
-| dummyDeployment.affinity | object | {} | sets affinity rules on the  deployment |
-| dummyDeployment.enabled | bool | false | enable the dummy deployment |
-| dummyDeployment.image.pullPolicy | string | IfNotPresent | When to pull the container image |
-| dummyDeployment.image.repository | string | k8s.gcr.io/pause | Container image to deploy |
-| dummyDeployment.image.tag | string | 3.4.1 | sets the image tag to use |
-| dummyDeployment.nodeSelector | object | {} | sets a nodeSelector on the  deployment |
-| dummyDeployment.podAnnotations | object | {} | sets Pod annotations on the deployment |
-| dummyDeployment.resources | object | {} | sets resources like limits and requests on the  deployment |
-| dummyDeployment.tolerations | list | [] | sets tolerations on the  deployment |
+| dummyDeployment.affinity | object | `{}` | sets affinity rules on the  deployment |
+| dummyDeployment.enabled | bool | `false` | enable the dummy deployment |
+| dummyDeployment.image.pullPolicy | string | `"IfNotPresent"` | When to pull the container image |
+| dummyDeployment.image.repository | string | `"k8s.gcr.io/pause"` | Container image to deploy |
+| dummyDeployment.image.tag | string | `"3.4.1"` | sets the image tag to use |
+| dummyDeployment.nodeSelector | object | `{}` | sets a nodeSelector on the  deployment |
+| dummyDeployment.podAnnotations | object | `{}` | sets Pod annotations on the deployment |
+| dummyDeployment.resources | object | `{}` | sets resources like limits and requests on the  deployment |
+| dummyDeployment.tolerations | list | `[]` | sets tolerations on the  deployment |
 | parameters | object | `{}` | parameters for provider |
 | parametersTpl | string | raw values from `parameters` | template to generate parameters |
 | provider | string | `nil` | specify secret provider |
--- a/charts/csi-secret-provider-class/templates/_helpers.tpl
+++ b/charts/csi-secret-provider-class/templates/_helpers.tpl
@@ -1,41 +1,10 @@
 {{/* vim: set filetype=mustache: */}}
-{{/*
-Expand the name of the chart.
-*/}}
-{{- define "csi-secret-provider-class.name" -}}
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
-{{- end }}
-
-{{/*
-Create a default fully qualified app name.
-We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
-If release name contains chart name it will be used as a full name.
-*/}}
-{{- define "csi-secret-provider-class.fullname" -}}
-{{- if .Values.fullnameOverride }}
-{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- $name := default .Chart.Name .Values.nameOverride }}
-{{- if contains $name .Release.Name }}
-{{- .Release.Name | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
-{{- end }}
-{{- end }}
-{{- end }}
-
-{{/*
-Create chart name and version as used by the chart label.
-*/}}
-{{- define "csi-secret-provider-class.chart" -}}
-{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
-{{- end }}
 
 {{/*
 Common labels
 */}}
 {{- define "csi-secret-provider-class.labels" -}}
-helm.sh/chart: {{ include "csi-secret-provider-class.chart" . }}
+helm.sh/chart: {{ include "common.chartref" . }}
 {{ include "csi-secret-provider-class.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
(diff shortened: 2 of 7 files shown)


charts/csi-secret-provider-class/templates/_helpers.tpl
```diff
@@ -0,0 +1,63 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "csi-secret-provider-class.name" -}}
```
hairmare, line 5: Most of these are covered by `common` and are not used from here.

charts/csi-secret-provider-class/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if .Values.dummyDeployment.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  {{ template "common.metadata" $ }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "csi-secret-provider-class.selectorLabels" . | nindent 6 }}
```
hairmare, line 9: There might be something generic in `common` to replace this but I'd need to check first.

charts/csi-secret-provider-class/values.yaml
```diff
@@ -14,3 +14,34 @@ secretObjects: []
 # secretObjectsTpl -- template to generate secretObjects
 # @default -- raw values from `secretObjects`
 secretObjectsTpl: "{{ .Values.secretObjects | toYaml }}"
+
+# Creates a dummy deployment which uses the csiVolue. This is a hack, in case your Deployment in case your App/Chart isn't enabled to define extraVolumes
```
hairmare, line 18:
```suggestion
# Creates a dummy Deployment which uses the csiVolume. This is a hack, in case your Chart doesn't support defining extraVolumes
```

charts/csi-secret-provider-class/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if .Values.dummyDeployment.enabled }}
+apiVersion: apps/v1
+kind: Deployment
```
hairmare, line 3: Could this be one of the rare edge cases where we consider directly managing a Pod rather than using a Deployment?

charts/csi-secret-provider-class/templates/deployment.yaml
eyenx: found nothing. This is why I created `_helpers.tpl` in the first place.

charts/csi-secret-provider-class/templates/deployment.yaml
eyenx: We just have to think about the Deploymentstrategy. What does need to happen on changes? Remove the Pod and recreate it? The secret gets removed for a brief short of time. This is why I thought about using the Deployment, but haven't thought about Rollingupdate and so on yet.

tongpu:
ELI5: This is required to run once to ensure that the secret is being populated? Other possible hacks I could think about is just running a Job as a Helm hook. But I'm unsure if we really need a Helm chart for everything, even hacks like that?

eyenx:
> ELI5: This is required to run once to ensure that the secret is being populated? Other possible hacks I could think about is just running a Job as a Helm hook. But I'm unsure if we really need a Helm chart for everything, even hacks like that?

The pod needs to run always.

charts/csi-secret-provider-class/templates/deployment.yaml
hairmare: i took a look at the original incubator upstream as well and could not find anything there either

charts/csi-secret-provider-class/templates/_helpers.tpl
```diff
@@ -0,0 +1,63 @@
...
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "csi-secret-provider-class.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "csi-secret-provider-class.name" . }}
```
hairmare, line 50: I think this could use something based on `common.name`.

charts/csi-secret-provider-class/templates/_helpers.tpl
```diff
@@ -0,0 +1,63 @@
...
+*/}}
+{{- define "csi-secret-provider-class.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "csi-secret-provider-class.labels" -}}
+helm.sh/chart: {{ include "csi-secret-provider-class.chart" . }}
```
hairmare, line 38: Might be replaceable with `common.chartref`

charts/csi-secret-provider-class/templates/_helpers.tpl
```diff
@@ -0,0 +1,63 @@
...
+*/}}
+{{- define "csi-secret-provider-class.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "csi-secret-provider-class.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "csi-secret-provider-class.serviceAccountName" -}}
```
hairmare, line 57: This one isn't used anywhere

hairmare (commented):
some more food for thought regarding using common in helpers

charts/csi-secret-provider-class/templates/deployment.yaml
hairmare: You are right, the changes will be much more straight forward if it's a Deployment.

hairmare:
> This is required to run once to ensure that the secret is being populated? Other possible hacks I could think about is just running a Job as a Helm hook

A "run once" solution might be done with [argocd-vault-plugin](https://github.com/IBM/argocd-vault-plugin) but this aims at tricking the csi bits into continually managing the secret.

charts/csi-secret-provider-class/values.yaml
```diff
@@ -14,3 +14,34 @@ secretObjects: []
 # secretObjectsTpl -- template to generate secretObjects
 # @default -- raw values from `secretObjects`
 secretObjectsTpl: "{{ .Values.secretObjects | toYaml }}"
+
+# Creates a dummy Deployment which uses the csiVolume. This is a hack, in case your Chart doesn't support defining extraVolumes
```
hairmare, line 18: Maybe we should add another strongly worded sentence here that tells people not to use this unless they tried fixing the underlying issue and can prove this with a PR or support ticket for the affected chart.

charts/csi-secret-provider-class/templates/_helpers.tpl
eyenx: 11d0f3e759ba214811f4d382bfbac26083874c9b

charts/csi-secret-provider-class/templates/_helpers.tpl
eyenx: 11d0f3e759ba214811f4d382bfbac26083874c9b

charts/csi-secret-provider-class/templates/_helpers.tpl
eyenx: 11d0f3e759ba214811f4d382bfbac26083874c9b

charts/csi-secret-provider-class/templates/_helpers.tpl
eyenx: 11d0f3e759ba214811f4d382bfbac26083874c9b

charts/csi-secret-provider-class/templates/_helpers.tpl
eyenx: 11d0f3e759ba214811f4d382bfbac26083874c9b

charts/csi-secret-provider-class/values.yaml
eyenx:
```suggestion
# Creates a dummy Deployment which uses the csiVolume. This is a hack, in case your Chart doesn't support defining extraVolumes
# DO NOT USE THIS unless strictly needed. Always try to use extraVolume Bindings inside your already existing Chart or application to make csi-secret-store manage the secret
```

eyenx: Something like this?

charts/csi-secret-provider-class/values.yaml
```diff
@@ -14,3 +14,34 @@ secretObjects: []
...
+  enabled: false
+  image:
+    # dummyDeployment.image.repository -- Container image to deploy
+    # @default -- k8s.gcr.io/pause
+    repository: k8s.gcr.io/pause
+    # dummyDeployment.image.tag -- sets the image tag to use
+    # @default -- 3.4.1
+    tag: 3.4.1
+    # dummyDeployment.image.pullPolicy -- When to pull the container image
+    # @defualt -- ifNotPresent
```
tongpu, line 31:
```suggestion
    # @default -- ifNotPresent
```

charts/csi-secret-provider-class/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+      {{- with .Values.dummyDeployment.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.dummyDeployment.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      volumes:
+        - name: csi-secret-store-volume
```
tongpu, line 40: Is it sufficient to define the volume in `volumes` without defining a `volumeMount` in any container?

tongpu (commented):
This discussion is probably beyond the scope of the pull request, but what would happen if this deployment is never scheduled because of resource exhaustion?
The secret would never be created and the other application without the `extraVolumes` will also never be scheduled, because it's waiting for a secret to be available in the Kubernetes API?

hairmare:
> This discussion is probably beyond the scope of the pull request, but what would happen if this deployment is never scheduled because of resource exhaustion?

It's a tiny pause container, if we run into this then we have large issues elsewhere. We should probably test how it reacts in those scenarios at some point though. I would not be surprised if we find room for improvement in these areas of the CSI stack.

charts/csi-secret-provider-class/templates/deployment.yaml
eyenx: yes.

eyenx:
> > This is required to run once to ensure that the secret is being populated? Other possible hacks I could think about is just running a Job as a Helm hook
> 
> A "run once" solution might be done with [argocd-vault-plugin](https://github.com/IBM/argocd-vault-plugin) but this aims at tricking the csi bits into continually managing the secret.

lol we might keep an eye open for that plugin.

charts/csi-secret-provider-class/values.yaml
hairmare: yeah & with an additional mention of the upstream work that we expect you to put in before you are allowed to use it

charts/csi-secret-provider-class/templates/deployment.yaml
hairmare: s/yes/maybe/ 🤨

eyenx: AH!

eyenx: But not by the "dummyDeployment" which is NOT using the Secret right away.

charts/csi-secret-provider-class/values.yaml
eyenx: Better?

hairmare: yes, thanks!

charts/csi-secret-provider-class/templates/deployment.yaml
hairmare: we might want to investigate this further, i can't really explain what has been happeining yet but i can take another deep dive into the implementation to figure it out if we don't identify a pattern.

charts/csi-secret-provider-class/values.yaml
hairmare: Did you end up changing this?

eyenx: Just now.

eyenx: Now checks are failing... ngg

charts/csi-secret-provider-class/values.yaml
```diff
@@ -14,3 +14,36 @@ secretObjects: []
 # secretObjectsTpl -- template to generate secretObjects
 # @default -- raw values from `secretObjects`
 secretObjectsTpl: "{{ .Values.secretObjects | toYaml }}"
+
+# Creates a dummy Deployment which uses the csiVolume. This is a hack, in case your Chart doesn't support defining extraVolumes
+# DISCLAIMER: DO NOT USE THIS unless as last resort.
```
eyenx, line 19:
```suggestion
# DISCLAIMER: DO NOT USE THIS unless as last resort.
```

charts/csi-secret-provider-class/values.yaml
eyenx: Checks are passing @hairmare

charts/csi-secret-provider-class/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
+{{- if .Values.dummyDeployment.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  {{ template "common.metadata" $ }}
```
hairmare, line 5: The nesting seems wrong here:
```suggestion
{{ template "common.metadata" . }}
```
Also, there is no reason to pass `$` when `.` contains the same content.

hairmare (changes requested):
Apart from the nesting error below you should also add `dummyDeployment.enabled=true` to `ci/default-values.yaml` for ensuring that the test generates a deployment.

charts/csi-secret-provider-class/templates/deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+        {{- include "csi-secret-provider-class.selectorLabels" . | nindent 8 }}
+    spec:
+      containers:
+        - name: dummy
+          image: {{ .Values.dummyDeployment.image.repository }}:{{ .Values.dummyDeployment.image.tag }}
+          imagePullPolicy: {{ .Values.dummyDeployment.image.pullPolicy }}
+          {{- with .Values.dummyDeployment.resources }}
+          resources:
+            {{- toYaml .Values.dummyDeployment.resources | nindent 12 }}
+          {{- end }}
```
hairmare, line 25: Use `.` when using `with`.
```suggestion
          {{- with .Values.dummyDeployment.resources }}
          resources:
            {{- toYaml .| nindent 12 }}
          {{- end }}
```

charts/csi-secret-provider-class/README.md
```diff
@@ -17,12 +17,21 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
...
-| https://charts.adfinis.com | common | 0.0.5 |
+| https://charts.adfinis.com | common | 0.0.6 |
 
 ## Values
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
+| dummyDeployment.affinity | object | {} | sets affinity rules on the  deployment |
+| dummyDeployment.enabled | bool | false | enable the dummy deployment |
+| dummyDeployment.image.pullPolicy | string | ifNotPresent | When to pull the container image |
```
hairmare, line 28:
```suggestion
| dummyDeployment.image.pullPolicy | string | IfNotPresent | When to pull the container image |
```

charts/csi-secret-provider-class/values.yaml
```diff
@@ -14,3 +14,36 @@ secretObjects: []
...
+  image:
+    # dummyDeployment.image.repository -- Container image to deploy
+    # @default -- k8s.gcr.io/pause
+    repository: k8s.gcr.io/pause
+    # dummyDeployment.image.tag -- sets the image tag to use
+    # @default -- 3.4.1
+    tag: 3.4.1
+    # dummyDeployment.image.pullPolicy -- When to pull the container image
+    # @default -- ifNotPresent
+    pullPolicy: ifNotPresent
```
hairmare, line 34:
```suggestion
    pullPolicy: IfNotPresent
```
Also, don't duplicate `@default` if it matches the actual value!

hairmare (changes requested):
Found the typo!

charts/csi-secret-provider-class/values.yaml
```diff
@@ -14,3 +14,36 @@ secretObjects: []
...
+  # dummyDeployment.enabled -- enable the dummy deployment
+  # @default -- false
+  enabled: false
+  image:
+    # dummyDeployment.image.repository -- Container image to deploy
+    # @default -- k8s.gcr.io/pause
+    repository: k8s.gcr.io/pause
+    # dummyDeployment.image.tag -- sets the image tag to use
+    # @default -- 3.4.1
+    tag: 3.4.1
```
hairmare, line 31:
```suggestion
  enabled: false
  image:
    # dummyDeployment.image.repository -- Container image to deploy
    repository: k8s.gcr.io/pause
    # dummyDeployment.image.tag -- sets the image tag to use
    tag: 3.4.1
```

charts/csi-secret-provider-class/values.yaml
```diff
@@ -14,3 +14,36 @@ secretObjects: []
...
+  nodeSelector: {}
+  # dummyDeployment.affinity -- sets affinity rules on the  deployment
+  # @default -- {}
+  affinity: {}
+  # dummyDeployment.tolerations -- sets tolerations on the  deployment
+  # @default -- []
+  tolerations: []
+  # dummyDeployment.resources -- sets resources like limits and requests on the  deployment
+  # @default -- {}
+  resources: {}
```
hairmare, line 49:
```suggestion
  podAnnotations: {}
  # dummyDeployment.nodeSelector -- sets a nodeSelector on the  deployment
  nodeSelector: {}
  # dummyDeployment.affinity -- sets affinity rules on the  deployment
  affinity: {}
  # dummyDeployment.tolerations -- sets tolerations on the  deployment
  tolerations: []
  # dummyDeployment.resources -- sets resources like limits and requests on the  deployment
  resources: {}
```

hairmare (changes requested):
Only use `@default` when it you need to override the generated default because it would produce illegible JSON output.

hairmare (approved):

PR #177, opened 2021-02-18
feat(cert-manager-issuers): initialize chart

### LINKED ISSUE #134: new chart: issuers and clusterissuers for cert-manager
**Is your feature request related to a problem? Please describe.**
We want to deploy Issuers and ClusterIssuers using the same tooling as eveything else.

**Describe the solution you'd like**
A Helm chart that can deploy multiple Issuers and/or ClusterIssuers.

**Describe alternatives you've considered**
We considered using `incubator/raw` but would like to ensure that we solve this in a way that let's us move forward without depending on `incubator` and without using `raw` which circumvents a lot of Helm to get stuff done.

**Affected chart**
None, this would be a new Chart.

**Additional context**
This has been discussed by cert-manager and they recommend deploying Issuers and ClusterIssuers using it's own manifest/chart after installing cert-manager:
* https://github.com/jetstack/cert-manager/issues/2716
* https://github.com/jetstack/cert-manager/pull/2778

hairmare:
Fixes #134


charts/cert-manager-issuers/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v2
```
tongpu, line 1: I was literally looking for the missing `kind` for about minute before I figured out that this is not a Kubernetes object YAML. :D

charts/cert-manager-issuers/ci/default-values.yaml
```diff
@@ -0,0 +1,25 @@
+issuers:
```
tongpu, line 1: Would issuers not require a namespace?

hairmare: And Helm will put them in it's target namespace if you don't specify one in `issuers[].metadata.namespace`. I should add an example that specify a namespace though.

hairmare: see `charts/cert-manager-issuers/examples/vault-pki-issuers-in-namespaces.yaml`

charts/cert-manager-issuers/examples/letsencrypt-clusterissuers.yaml
```diff
@@ -0,0 +1,25 @@
+# Default Let's Encrypt ClusterIssuers
+#
+# Configures the common letsencrypt-staging and letsencrypt-prod ClusterIssuers
+# using a http01 solver that will work well with nginx-ingress. This is what
+# most small single-tenant clusters start with.
+
+_: &email user@example.com
```
eyenx, line 7: What is this? 🤔

hairmare: late-stage-yaml.png

A YAML anchor that is used so the string "user@example.com" is only once in the file. It uses `_` as a dummy key since it looks like the Go blank operator.

charts/cert-manager-issuers/examples/selfsigned-singleton-clusterissuer.yaml
```diff
@@ -0,0 +1,8 @@
+# Deploy a singleton selfSigned ClusterIssuer.
+#
+# You can leave out the name and the chart will use a generated name. Use this format
+# when you only want to deploy a single issuer per release of the chart.
+
+clusterissuers:
```
tongpu, line 6:
```suggestion
clusterIssuers:
```

charts/cert-manager-issuers/templates/NOTES.txt
```diff
@@ -0,0 +1,6 @@
+{{- if .Values.issuers }}
+- Deployed Issuers
```
tongpu, line 2: It would be nice if we could output a list of issuers and/or clusterIssuers that have been deployed by the chart, but I was not able to figure out quickly how the `cert-manager-issuers.metadata` is generating the name of e.g. the sole `selfSigned` clusterIssuer.

hairmare: I'm gonna hate myself for how the code in NOTES.txt looks once I implement this but I do see a need for it.

Given that we don't usually look at the NOTES output when we use Argo CD I am tending to want to implement this in a future update to the chart. I'm hoping that this future update will also take into account any real world issues we encounter when we start using this chart (so I don't need to refactor the notes stuff twice).

I'll think some more about this, if I figure out a somewhat DRY way to do this I'll update this PR.

charts/cert-manager-issuers/examples/selfsigned-singleton-clusterissuer.yaml
hairmare: fixed, thanks

charts/cert-manager-issuers/templates/NOTES.txt
hairmare: I refactored the helper and rewrote NOTES.txt

PR #178, opened 2021-02-23
feat(cert-manager-issuers): enabled flag for issuers

hairmare:
So we can conveniently disable single issuers without having to comment the whole issuer.


eyenx:
Why not setting enabled = true as default?

eyenx:
Ahh get it now

eyenx (approved):

PR #179, opened 2021-02-23
fix(templates/deployment.yaml): needs a volumeMount to work

eyenx:
In fact it needs a volumeMount to work. I don't know what I was talking about last week.


hairmare (approved):
¯\_(ツ)_/¯

tongpu:
> In fact it needs a volumeMount to work. I don't know what I was talking about last week.

I hate it when I'm right. :D

eyenx:
I hate it too when you're right.

PR #196, opened 2021-03-18
fix(chart-releaser): add bitnami repo to deps

eyenx:
# Description

Fixes the repo missing in the Chart Releaser action.

We should really think about having this managed somewhere else.

# Issues

release fails with: `Error: no repository definition for https://charts.bitnami.com/bitnami, https://charts.bitnami.com/bitnami`

# Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -29,7 +29,8 @@ jobs:
       - name: Add dependency chart repos
         run: |
           helm repo add adfinis https://charts.adfinis.com && \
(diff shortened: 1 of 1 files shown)


vmaillot (approved):
LGTM

hairmare:
Maybe we could replace it with a script that grabs the repos from `./hack/chart-testing/ct.yaml`? We'd stilll need to manage both `./hack/chart-testing/ct.yaml` and `./hack/chart-testing/ct-install.yaml` but it would be a start.

PR #199, opened 2021-03-18
chore(infra-apps): bump Argo CD chart to 1.17.*

hairmare:
# Description
Update to an Argo CD chart version that contains Arrgo CD 1.8.4.

Below are the changes to the chart, please see the [release notes](https://github.com/argoproj/argo-cd/releases) for details on the content of the Argo CD update.

## Features:
* enable support for lifecycle hooks on ArgoCD Server
* conditionally create the rbac configmap
* update to 1.8.4
* support for readOnlyRootFilesystem in dex
* Mount ephemeral emptyDir on repoServer for /tmp and /app/config/gpg/keys
* add self-heal-timeout as a template variable
* adds configEnabled flag to allow parent chart to own CM
* upgrade redis-ha to v4.10.4
* add spec for the status field of AppProject CRD
* add possibility for change NodePort if this is type of Service
* add annotations to the dex service account
* create roles instead of rolebindings when singleNamespace is true

## Fixes:
* replace hard coded .selector.matchLabels (`-application-controller` suffix is now templates via `{{ .Values.controller.name }}`)

# Issues

# Checklist

### DIFF
--- a/charts/infra-apps/Chart.yaml
+++ b/charts/infra-apps/Chart.yaml
@@ -3,8 +3,8 @@ name: infra-apps
 description: Argo CD app-of-apps config for infrastructure components
 type: application
 # version and appVersion are in sync in this chart!
-version: 0.9.2
-appVersion: 0.9.2
+version: 0.9.3
+appVersion: 0.9.3
 home: https://github.com/adfinis-sygroup/helm-charts/tree/master/charts/infra-apps
 sources:
   - https://github.com/adfinis-sygroup/helm-charts
--- a/charts/infra-apps/Chart.yaml
+++ b/charts/infra-apps/Chart.yaml
@@ -3,8 +3,8 @@ name: infra-apps
 description: Argo CD app-of-apps config for infrastructure components
 type: application
 # version and appVersion are in sync in this chart!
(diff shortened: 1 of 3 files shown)


charts/infra-apps/Chart.yaml
```diff
@@ -3,8 +3,8 @@ name: infra-apps
 description: Argo CD app-of-apps config for infrastructure components
 type: application
 # version and appVersion are in sync in this chart!
-version: 0.9.2
-appVersion: 0.9.2
+version: 0.9.3
+appVersion: 0.9.3
```
tongpu, line 7: What's our policy regarding version bumps in the chart? Up to now I thought we're bumping the minor version of the chart when we increase the minor version of any subcharts?

hairmare: This should have absolutely been a bump at the minor level!

tongpu (approved):
LGTM

charts/infra-apps/Chart.yaml
eyenx: Do we want to write that down somewhere in our documentation?

PR #200, opened 2021-03-19
chore(vault-csi-provider): update provider to 0.0.7 and dep to 0.0.20

eyenx:
# Description

* Updates our vault-csi-provider to use image tag 0.0.7 and adds default resources, as in the deployment example [changelog](https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/releases/tag/0.0.7)
* Updates the dependency of secrets-store-csi-driver to 0.0.20 [changelog](https://github.com/kubernetes-sigs/secrets-store-csi-driver/releases/tag/v0.0.20)

# Issues

n/a

# Checklist

### DIFF
--- a/charts/vault-csi-provider/README.md
+++ b/charts/vault-csi-provider/README.md
@@ -21,7 +21,7 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
 |-----|------|---------|-------------|
 | csiProviderPath | string | `"/etc/kubernetes/secrets-store-csi-providers"` | set the path where the secrets-store-csi-provider gets installed on the node |
 | image | object | `{"pullPolicy":"IfNotPresent","repository":"hashicorp/secrets-store-csi-driver-provider-vault","tag":"0.0.7"}` | specifies the image to use for the secrets-store-csi-driver |
-| resources | object | `{"limits":{"cpu":"50m","memory":"100Mi"},"requests":{"cpu":"50m","memory":"100Mi"}}` | Set the limits and requests on vault csi-driver-provider pod resources |
+| resources | object | `{}` | Set the limits and requests on vault csi-driver-provider pod resources |
 | secretsStoreCsiDriver.enabled | bool | `true` | specifies wether or not the secrets-store-csi-driver dependency gets installed |
 
 ## About this chart
--- a/charts/vault-csi-provider/README.md
+++ b/charts/vault-csi-provider/README.md
@@ -1,6 +1,6 @@
 # vault-csi-provider
 
-![Version: 0.2.2](https://img.shields.io/badge/Version-0.2.2-informational?style=flat-square) ![AppVersion: 0.0.6](https://img.shields.io/badge/AppVersion-0.0.6-informational?style=flat-square)
+![Version: 0.2.3](https://img.shields.io/badge/Version-0.2.3-informational?style=flat-square) ![AppVersion: 0.0.7](https://img.shields.io/badge/AppVersion-0.0.7-informational?style=flat-square)
 
 A helm chart to install the vault secrets-store-csi-driver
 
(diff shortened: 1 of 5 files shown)


pree (approved):
LGTM

charts/vault-csi-provider/templates/daemonset.yaml
```diff
@@ -30,3 +30,5 @@ spec:
         - name: providervol
           hostPath:
             path: {{ .Values.csiProviderPath }}
+      nodeSelector:
+        beta.kubernetes.io/os: linux
```
tongpu, line 34: Is Karbon still using this with a beta prefix? `beta.kubernetes.io/os` is deprecated according to the [docs](https://kubernetes.io/docs/reference/labels-annotations-taints/) and `kubernetes.io/os` should be used instead.
```suggestion
        kubernetes.io/os: linux
```

tongpu (commented):
Please double check regarding the `kubernetes.io/os` label.

charts/vault-csi-provider/templates/daemonset.yaml
pree: It's the same on the upstream secrets-store-csi-driver-provider-vault: https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/blob/master/deployment/provider-vault-installer.yaml#L72

pree: `beta.` was deprecated in 1.14 and the csi-provider needs at least 1.15 - I think we can safely use `kubernetes.io/os`

charts/vault-csi-provider/README.md
```diff
@@ -13,15 +13,15 @@ This chart is maintained by [Adfinis](https://adfinis.com/?pk_campaign=github&pk
...
 
 ## Values
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
 | csiProviderPath | string | `"/etc/kubernetes/secrets-store-csi-providers"` | set the path where the secrets-store-csi-provider gets installed on the node |
-| image | object | `{"pullPolicy":"IfNotPresent","repository":"hashicorp/secrets-store-csi-driver-provider-vault","tag":"0.0.6"}` | specifies the image to use for the secrets-store-csi-driver |
-| resources | object | `{}` | Set the limits and requests on vault csi-driver-provider pod resources |
+| image | object | `{"pullPolicy":"IfNotPresent","repository":"hashicorp/secrets-store-csi-driver-provider-vault","tag":"0.0.7"}` | specifies the image to use for the secrets-store-csi-driver |
+| resources | object | `{"limits":{"cpu":"50m","memory":"100Mi"},"requests":{"cpu":"50m","memory":"100Mi"}}` | Set the limits and requests on vault csi-driver-provider pod resources |
```
hairmare, line 24: Can you update the doc so they don't contain extremely long JSON strings?

charts/vault-csi-provider/templates/daemonset.yaml
hairmare: In any case (independant of if there is a beta beta prefix or not), this nodeSelector should be confgured via values and not be a hardcoded magic string in the Daemonset.

charts/vault-csi-provider/values.yaml
```diff
@@ -6,11 +6,17 @@
...
 image:
   repository: hashicorp/secrets-store-csi-driver-provider-vault
-  tag: 0.0.6
+  tag: 0.0.7
   pullPolicy: IfNotPresent
 
 # resources -- Set the limits and requests on vault csi-driver-provider pod resources
-resources: {}
+resources:
+  requests:
```
hairmare, line 14: Why deviate from the default of leaving this empty? I can't find a mention of leaving it empty in the best-practices but an empty chart made with `helm create` has this info:
>   We usually recommend not to specify default resources and to leave this as a conscious
   choice for the user. This also increases chances charts run on environments with little
   resources, such as Minikube. If you do want to specify resources, uncomment the following
   lines, adjust them as necessary, and remove the curly braces after 'resources:'.

IMO leaving this as a conscious choise to the user still holds true.

hairmare (changes requested):
There are a few magic strings that should not be hardcoded:

charts/vault-csi-provider/templates/daemonset.yaml
tongpu: Hardcoding a `nodeSelector` is acceptable when we know that the daemonset can't be executed on Windows nodes, but we should probably provide an option to extend the `nodeSelector` or configure our default in `values.yaml`.

eyenx: what has this to do  with Karbon?

charts/vault-csi-provider/values.yaml
eyenx: It's all taken from the default deployment/daemonset.yaml of the upstream repo. https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/blob/master/deployment/provider-vault-installer.yaml

eyenx: Would you suggest discarding their upstream input and just leaving this blank?

charts/vault-csi-provider/templates/daemonset.yaml
eyenx: It's also just taken from upstream here.

eyenx: https://github.com/hashicorp/secrets-store-csi-driver-provider-vault/blob/master/deployment/provider-vault-installer.yaml

charts/vault-csi-provider/values.yaml
sinicis: if some resources are required that the pod can work then it makes sense to define the minimal `requests`. `limits` are not needed IMHO

eyenx: According to theyr upstream chart which will be released soon, they also set the resources to nil. I will look into that later this month or in April so we can depreacate mantaining `vault-auth` and `vault-csi-provider` ourselfes.

charts/vault-csi-provider/README.md
eyenx: Removed.

charts/vault-csi-provider/values.yaml
eyenx: Removed and set to nil

sinicis (approved):

tongpu (approved):

PR #235, opened 2021-04-21
refactor(vaultauth/vaultcsiprovider): removal

eyenx:
# Description

BREAKING CHANGE: vaultAuth and vaultCSIProvider won't be part of the security-apps anymore starting
with version 1.0.0

# Issues
n/a

# Checklist


charts/security-apps/Chart.yaml
```diff
@@ -3,8 +3,8 @@ name: security-apps
 description: Argo CD app-of-apps config for security applications
 type: application
 # version and appVersion are in sync in this chart!
-version: 0.13.0
-appVersion: 0.13.0
+version: 1.0.0
+appVersion: 1.0.0
```
hairmare, line 7:
```suggestion
version: 0.14.0
appVersion: 0.14.0
```

breaking change rule for major bumps only applies to 1.x and higher versions.

this chart is still using 0.x for that exact purpose and should not get released as stable 1.x until we are sure that we have tried and tested all aspects of the chart and it's contents.

eyenx: Wow, didn't know that. TIL. Thanks

hairmare (approved):
LGTM

had there been any nicer examples in the `examples/` subdir i might have asked you to preserve their contents into the main vault example (which most likely also isn't up to scratch anyway)

PR #245, opened 2021-05-05
chore(timed): bump v1.1.4 and add new flags

sbor23:
# Description

Bump timed backend to v1.1.4

# Issues

# Checklist

* [ ] If I updated a dependency tool, or app, this PR contains a short summary of the changes I'm pulling

### DIFF
--- a/charts/timed/Chart.yaml
+++ b/charts/timed/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.6.3
-appVersion: v1.1.3
+version: 0.7.0
+appVersion: v1.1.4
 keywords:
   - timed
   - tracking
--- a/charts/timed/templates/configmap-backend.yaml
+++ b/charts/timed/templates/configmap-backend.yaml
@@ -59,9 +59,12 @@ data:
   DJANGO_REDMINE_HTACCESS_USER: {{ .Values.redmine.htaccessUser | quote }}
   DJANGO_REDMINE_SPENTHOURS_FIELD: {{ .Values.redmine.spenthoursField | quote }}
 {{- end }}
-  UWSGI_MAX_REQUESTS: {{ .Values.backend.settings.uwsgi.maxRequests | quote }}
-  UWSGI_HARAKIRI: {{ .Values.backend.settings.uwsgi.harakiri | quote }}
-  UWSGI_PROCESSES: {{ .Values.backend.settings.uwsgi.processes | quote }}
+  GUNICORN_WORKERS: {{ .Values.backend.settings.gunicorn.workers | quote }}
+  GUNICORN_CMD_ARGS: {{ .Values.backend.settings.gunicorn.cmdArgs | quote }}
 {{- if .Values.backend.settings.admins }}
   DJANGO_ADMINS: {{ join "," .Values.backend.settings.admins | quote }}
+{{- end }}
+{{- if .Values.backend.settings.sentry.enabled }}
+  DJANGO_SENTRY_TRACES_SAMPLE_RATE: {{ .Values.backend.settings.sentry.tracesSampleRate | quote }}
+  DJANGO_SENTRY_SEND_DEFAULT_PII: {{ .Values.backend.settings.sentry.sendDefaultPii | quote }}
 {{- end }}
--- a/charts/timed/values.yaml
+++ b/charts/timed/values.yaml
@@ -74,13 +74,21 @@ backend:
     workReportPath: /etc/workreport
     # backend.settings.admins -- Django administrators, example: Jon Doe <jon.doe@example.com>
     admins: []
-    uwsgi:
-      # backend.settings.uwsgi.maxRequests -- uWSGI max requests
-      maxRequests: 2000
-      # backend.settings.uwsgi.harakiri -- uWSGI harakiri (request timeout)
-      harakiri: 60
-      # backend.settings.uwsgi.processes -- uWSGI number of worker processes
-      processes: 16
+    gunicorn:
+      # backend.settings.gunicorn.workers -- Number of gunicorn workers
+      workers: 8
+      # backend.settings.gunicorn.cmdArgs -- gunicorn additional arguments
+      cmdArgs: ""
+    sentry:
+      # backend.settings.sentry.enabled -- Enable Sentry integration
+      enabled: false
+      # backend.settings.sentry.dsn -- Sentry DSN for error reporting. Set to enable Sentry integration
+      dsn: ~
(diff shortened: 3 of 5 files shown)


charts/timed/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.6.3
-appVersion: v1.1.3
+version: 0.6.4
```
tongpu, line 5: With regards to the Helm chart configuration this would need to be a minor version bump, because we're removing configuration options and replacing them with new ones.
```suggestion
version: 0.7.0
```

charts/timed/values.yaml
```diff
@@ -307,13 +305,19 @@ redmine:
   url: redmine.example.com
   # redmine.apiKey -- Redmine API Key
   apiKey: ""
-  # redmine.htaccessUser -- Redmine htaccess user
-  htaccessUser: ""
-  # redmine.htaccessPassword -- Redmine htaccess password
```
tongpu, line 312: I guess these settings are specific to our environment and not needed anymore since we migrated Redmine away from the old environment?

tongpu (changes requested):
These changes require a minor version bump of the Helm chart, otherwise we might break existing setups.

charts/timed/values.yaml
sbor23: These settings were actually useless from the beginning, hence they were dropped. The `apiKey` suffices for authentication with redmine.
See https://github.com/adfinis-sygroup/timed-backend/pull/733

sbor23:
For some reason the backend pod can't connect to the db and that's something I can't reproduce locally (using the dev-setup with docker-compose unfortunately...). So I don't think this was introduced by gunicorn (https://github.com/adfinis-sygroup/timed-backend/pull/734)

charts/timed/Chart.yaml
sbor23: done

charts/timed/templates/configmap-backend.yaml
```diff
@@ -59,9 +59,13 @@ data:
...
-  UWSGI_MAX_REQUESTS: {{ .Values.backend.settings.uwsgi.maxRequests | quote }}
-  UWSGI_HARAKIRI: {{ .Values.backend.settings.uwsgi.harakiri | quote }}
-  UWSGI_PROCESSES: {{ .Values.backend.settings.uwsgi.processes | quote }}
+  GUNICORN_WORKERS: {{ .Values.backend.settings.gunicorn.workers | quote }}
+  GUNICORN_CMD_ARGS: {{ .Values.backend.settings.gunicorn.cmdArgs | quote }}
 {{- if .Values.backend.settings.admins }}
   DJANGO_ADMINS: {{ join "," .Values.backend.settings.admins | quote }}
 {{- end }}
+{{- if .Values.sentry.enabled }}
+  DJANGO_SENTRY_DSN: {{ .Values.sentry.dsn | quote }}
```
hairmare, line 68: This env var should be in `secrets-backend.yaml` and not in the configmap.

charts/timed/values.yaml
```diff
@@ -307,13 +305,19 @@ redmine:
...
-  htaccessPassword: ""
   # redmine.spenthoursField -- Spent hours field id on Redmine
   spenthoursField: 0
 
+sentry:
+  # sentry.enabled -- Enable Sentry integration
+  enabled: false
+  # sentry.dsn -- Sentry DSN for error reporting. Set to enable Sentry integration
+  dsn: sentry.example.com
```
hairmare, line 315: using `nil` forces the user to overwrite this:
```suggestion
  dsn: ~
```

charts/timed/values.yaml
```diff
@@ -307,13 +305,19 @@ redmine:
...
+  # sentry.tracesSampleRate -- Sentry trace sample rate, Set 1.0 to capture 100% of transactions
+  tracesSampleRate: 1.0
+  # sentry.sendDefaultPii -- Associate users to errors in Sentry
+  sendDefaultPii: true
```
hairmare, line 319: Doesn't this need to be Python `True` (and quoted)?
```suggestion
  sendDefaultPii: "True"
```

charts/timed/Chart.lock
```diff
@@ -3,4 +3,4 @@ dependencies:
   repository: https://charts.bitnami.com/bitnami
   version: 9.1.4
 digest: sha256:f85f25319e458a5dcfc6e5a090ffe452b19b5104cbc24c2b917d799edb49eebf
-generated: "2021-04-26T15:10:58.064800819+02:00"
+generated: "2021-05-05T08:48:13.729525413+02:00"
```
hairmare, line 6: something is fishy about this change, when you don't change deps this does not need an update

charts/timed/values.yaml
```diff
@@ -307,13 +305,19 @@ redmine:
...
   # redmine.apiKey -- Redmine API Key
   apiKey: ""
-  # redmine.htaccessUser -- Redmine htaccess user
-  htaccessUser: ""
-  # redmine.htaccessPassword -- Redmine htaccess password
-  htaccessPassword: ""
   # redmine.spenthoursField -- Spent hours field id on Redmine
   spenthoursField: 0
 
+sentry:
```
hairmare, line 311: The sentry settings are specifig to the backend:
```suggestion
backend:
  sentry:
```

hairmare (changes requested):

charts/timed/values.yaml
sbor23: Hmm there's more settings like `redmine` which is its own object, so I figured to keep it separate too.
I can integrate them both into `backend.settings`, sure

charts/timed/values.yaml
sbor23: Ahh this gets passed 1:1 as a env var, sure then.

hairmare:
> For some reason the backend pod can't connect to the db and that's something I can't reproduce locally (using the dev-setup with docker-compose unfortunately...). So I don't think this was introduced by gunicorn ([adfinis-sygroup/timed-backend#734](https://github.com/adfinis-sygroup/timed-backend/pull/734))

Looks like the issue is that the frontend pod won't start due to it trying to pull v1.1.4... You need to set `frontend.image.tag=v.1.1.3` since there is only a backend for 1.1.4.

charts/timed/values.yaml
hairmare: It matters here because we might also add an integration for the frontend and that will need it's own DSN. redmine and others don't have this distinction.

hairmare (approved):
LGTM

tongpu (approved):
LGTM

PR #287, opened 2021-06-22
Add openshift-etcd-backup chart

vmaillot:
# Description

Add `openshift-etcd-backup` chart to simply backup your etcd data in OpenShift clusters.

# Issues

# Checklist


charts/openshift-etcd-backup/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
+name: openshift-etcd-backup
+description: Chart for openshfit-etcd-backup solution
+type: application
+version: 0.1.0
+appVersion: 1.0.0
```
tongpu, line 6: There's not a single tagged release available on [adfinis-sygroup/openshift-etcd-backup](https://github.com/adfinis-sygroup/openshift-etcd-backup), so I'm not sure if we should use 1.0.0 as the appVersion.

charts/openshift-etcd-backup/templates/configmap.yaml
```diff
@@ -0,0 +1,87 @@
...
+  labels:
+    {{- include "openshift-etcd-backup.labels" . | nindent 4 }}
+data:
+  OCP_BACKUP_SUBDIR: {{ .Values.backup.subdir | quote }}
+  OCP_BACKUP_DIRNAME: {{ .Values.backup.dirname | quote }}
+  OCP_BACKUP_EXPIRE_TYPE: {{ .Values.backup.expiretype | quote }}
+  OCP_BACKUP_KEEP_DAYS: {{ .Values.backup.keepdays | quote }}
+  OCP_BACKUP_KEEP_COUNT: {{ .Values.backup.keepcount | quote }}
+  backup.script: |
+    #!/bin/sh
```
hairmare, line 14: please don't include a copy paste of the original upstream source for this script as a configmap

charts/openshift-etcd-backup/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+    path: "/opt/etcd-backups"
+  provisionned:
+    # Enable provisionned backend storage with default or overrided storageClass
+    enabled: false
+    storageClass: ""
+
+image:
+  # change repository value below with the result of the following command
+  # `oc adm release info --image-for=tools | cut -d: -f1`
+  repository: quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256
```
hairmare, line 46: we should replace this with an image of our own that already contains the script

charts/openshift-etcd-backup/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  # change tag value below with the result of the following command
+  # `oc adm release info --image-for=tools | cut -d: -f2`
+  tag: ""
+
+nameOverride: ""
+fullnameOverride: "etcd-backup"
```
hairmare, line 54: should be empty per default
```suggestion
fullnameOverride: ""
```

charts/openshift-etcd-backup/values.yaml
```diff
@@ -0,0 +1,92 @@
+# Default values for openshift-etcd-backup.
```
hairmare, line 1: Values in here are all missing [helm-docs](https://github.com/norwoodj/helm-docs) style comments

charts/openshift-etcd-backup/templates/pv.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if .Values.persistence.nfs.enabled }}
+apiVersion: v1
+kind: PersistentVolume
```
hairmare, line 3: should this be a PVC? directly managing a PV seems non intuitive

charts/openshift-etcd-backup/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+    # NFS server name or IP
+    server: example.com
+    # NFS server path
+    path: "/etcd-backups"
+  hostPath:
+    # Enable hostPath
+    enabled: true
+    # hostPath existing path on host
+    path: "/opt/etcd-backups"
+  provisionned:
```
hairmare, line 38:
```suggestion
  provisioned:
```

charts/openshift-etcd-backup/templates/pv.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if .Values.persistence.nfs.enabled }}
```
tongpu, line 1: Instead of creating a PV you could probably configure the NFS volume directly in `spec.volumes` of the pod.
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#nfsvolumesource-v1-core

charts/openshift-etcd-backup/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
...
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end }}
+
+{{/*
+Selector labels
+*/}}
+{{- define "openshift-etcd-backup.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "openshift-etcd-backup.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
```
hairmare, line 51: it doesn't look like selectorLabels is being used anywhere exept labels:
```suggestion
{{- define "openshift-etcd-backup.labels" -}}
helm.sh/chart: {{ include "openshift-etcd-backup.chart" . }}
app.kubernetes.io/name: {{ include "openshift-etcd-backup.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
{{- end }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}
```

hairmare (dismissed):

charts/openshift-etcd-backup/values.yaml
tongpu, line 38: I would recommend to rename the variable from `provisionned` to `provisioning`. This would then also need to be updated in the templates.

```suggestion
  provisioning:
```

charts/openshift-etcd-backup/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+    # Enable provisionned backend storage with default or overrided storageClass
```
tongpu, line 39: Fix typo and improve wording
```suggestion
    # Enable provisioned backend storage with default or configured storageClass
```

tongpu (changes requested):
Quick first review. I would consider the change to remove the PV object the highest priority, even if the way I recommend it is not very well known.

charts/openshift-etcd-backup/templates/pv.yaml
tongpu: No, this should not be a PVC, because it directly configures a specific NFS volume. I'm not sure if my recommendation to directly configure the NFS volume in the pod spec under `spec.volumes` is any more intuitive, but in my opinion it would be the right way for static volumes that are specific to a pod.

charts/openshift-etcd-backup/Chart.yaml
vmaillot: what do you suggest? should we use the same as `version` then?

charts/openshift-etcd-backup/values.yaml
paraenggu: The idea here was to use an image which is always available, especially while running in restricted environments where the images are served by a mirror registry with limitations on what can be pulled-in.

README.md
```diff
@@ -39,6 +39,7 @@ Configure cert-manager Issuers and ClusterIssuers via Helm
 | [infra-apps](charts/infra-apps) | Argo CD app-of-apps config for infrastructure components | ![Version: 0.x](https://img.shields.io/badge/version-0.x-brightgreen) ![App version: 0.x](https://img.shields.io/badge/app%20version-0.x-brightgreen) |
 | [logging-apps](charts/logging-apps) | Argo CD app-of-apps config for logging applications | ![Version: 0.x](https://img.shields.io/badge/version-0.x-brightgreen) ![App version: 0.x](https://img.shields.io/badge/app%20version-0.x-brightgreen) |
 | [misc-apps](charts/misc-apps) | Argo CD app-of-apps config for miscellaneous small tools | ![Version: 0.x](https://img.shields.io/badge/version-0.x-brightgreen) ![App version: 0.x](https://img.shields.io/badge/app%20version-0.x-brightgreen) |
+| [openshift-etcd-backup](charts/openshift-etcd-backup) | Chart for openshfit-etcd-backup solution | ![Version: 0.x](https://img.shields.io/badge/version-0.x-brightgreen) ![App version: 1.0.x](https://img.shields.io/badge/app%20version-1.0.x-brightgreen) |
```
paraenggu, line 42: Typo: `openshfit` -> `openshift`

charts/openshift-etcd-backup/README.md
```diff
@@ -0,0 +1,67 @@
+# openshift-etcd-backup
+
+![Version: 0.1.0](https://img.shields.io/badge/Version-0.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.0.0](https://img.shields.io/badge/AppVersion-1.0.0-informational?style=flat-square)
+
+Chart for openshfit-etcd-backup solution
```
paraenggu, line 5: Typo: `openshfit`

charts/openshift-etcd-backup/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+    enabled: false
+    # NFS server name or IP
+    server: example.com
+    # NFS server path
+    path: "/etcd-backups"
+  hostPath:
+    # Enable hostPath
+    enabled: true
+    # hostPath existing path on host
+    path: "/opt/etcd-backups"
```
paraenggu, line 37: [`/opt/`](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s13.html) is probably not the best path here, use something below [`/mnt`](https://refspecs.linuxfoundation.org/FHS_3.0/fhs-3.0.html#mntMountPointForATemporarilyMount) (in case it's temporary) or something below [`/var/opt/`](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s12.html).

charts/openshift-etcd-backup/Chart.yaml
tongpu: We should release a version upstream and as @hairmare recommended build our own image based on the OpenShift ART image.

paraenggu: @tongpu see [https://github.com/adfinis-sygroup/helm-charts/pull/287/files#r656032602 #r656032602] for the reason behind the usage of the upstream image.

charts/openshift-etcd-backup/templates/pv.yaml
hairmare: either managing PVs directly feels like creating Pods directly or i don't understand what is going on here

vmaillot: it's all gone :D

charts/openshift-etcd-backup/Chart.yaml
hairmare: container images are for application delivery, configmaps are for configs

if need be we can push our image to quay so it will come from the same source as the image used upstream

paraenggu: > if need be we can push our image to quay so it will come from the same source as the image used upstream

No, the image originates from the internal mirror registry not from quay in this case.

charts/openshift-etcd-backup/templates/pv.yaml
vmaillot: cleaned!

charts/openshift-etcd-backup/templates/_helpers.tpl
vmaillot: thanks, it's done

charts/openshift-etcd-backup/values.yaml
vmaillot: thanks for the typo

README.md
vmaillot: updated!

charts/openshift-etcd-backup/values.yaml
vmaillot: thanks for the heads-up on that

charts/openshift-etcd-backup/templates/pv.yaml
paraenggu: @vmaillot why is it gone? 

A pre-created dedicated backup volume shall be used here. This volume mustn't be provisioned dynamically over a provisioner. The PV is needed in case of a restore, where one doesn't have a working cluster available. It is therefore crucial to know the exact volume mount information so that you can access the backup data on that volume independently.

vmaillot: because we can do it directly into the pod!
as @tongpu mentionned [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#nfsvolumesource-v1-core) you are
i guess we will have to discuss that subject

tongpu: @hairmare 
> either managing PVs directly feels like creating Pods directly or i don't understand what is going on here

Managing PVs directly doesn't come up too often, but it is a valid use-case for certain things.

@paraenggu 
> A pre-created dedicated backup volume shall be used here. This volume mustn't be provisioned dynamically over a provisioner. The PV is needed in case of a restore, where one doesn't have a working cluster available. It is therefore crucial to know the exact volume mount information so that you can access the backup data on that volume independently.

Question: The statement that the volume mustn't be provisioned dynamically is only valid in case we're using a static NFS mount correct?

The way I recommended to implement this is to to specify the NFS volume mount directly in the pod specification (under `spec.volumes`), without creating a PV at all. In my opinion there are a few benefits to configure it this way.

1. No one can accidentally delete the PV
2. The NFS configuration is directly configured in the Cronjob specification and it's clear that it should only be used for that.

Regarding restoring we would have to discuss the exact restore scenarios, because I can't yet imagine an etcd restore scenario where you have a working Kubernetes API to fetch the PV from.

charts/openshift-etcd-backup/Chart.yaml
tongpu: > container images are for application delivery, configmaps are for configs
> 
> if need be we can push our image to quay so it will come from the same source as the image used upstream

@hairmare This is image should™ always be present in the image cache on every OpenShift node, because it also used for some internal OpenShift shenanigans.

charts/openshift-etcd-backup/templates/pv.yaml
paraenggu: @tongpu 
> Question: The statement that the volume mustn't be provisioned dynamically is only valid in case we're using a static NFS mount correct?

Usually we will probably have an NFS share for the etcd-backup, but it could also be another NAS-like share.

> The way I recommended to implement this is to to specify the NFS volume mount directly in the pod specification (under `spec.volumes`), without creating a PV at all.

As long as we're able to specify the required GIDs that's fine, but aren't we limited then to NFS? @eni23 idea was that with a PV we get the flexibility to use whatever the customer provides.

> because I can't yet imagine an etcd restore scenario where you have a working Kubernetes API to fetch the PV from

Exactly, that's why we need to be able to access (mount) the share outside the cluster.

tongpu: > > Question: The statement that the volume mustn't be provisioned dynamically is only valid in case we're using a static NFS mount correct?
> 
> Usually we will probably have an NFS share for the etcd-backup, but it could also be another NAS-like share. However, it's never preferred to provision the volume dynamically, as we need it outside the cluster as well (where one doesn't know the internal volume path which was created during a dynamic provisioning in advance).

Yes, that part was clear for me.
 
> > The way I recommended to implement this is to to specify the NFS volume mount directly in the pod specification (under `spec.volumes`), without creating a PV at all.
> 
> As long as we're able to specify the [required GIDs](https://docs.openshift.com/container-platform/4.7/storage/persistent_storage/persistent-storage-nfs.html#storage-persistent-storage-nfs-group-ids_persistent-storage-nfs) that's fine,

If this is required to make it work, we need to add the ability to configure the Security Context to the chart, because I believe it's not yet implemented. This is not related to volume mounting at all.

> but aren't we limited then to NFS? @eni23 idea was that with a PV we get the flexibility to use whatever the customer provides.

With the way it was implemented creating the PV as part of the Helm chart we would have also been limited to NFS. The main questions is if we want to create the PV as part of the Helm chart or want to consume a preconfigured PV.

> > because I can't yet imagine an etcd restore scenario where you have a working Kubernetes API to fetch the PV from
> 
> Exactly, that's why we need to be able to access (mount) the share outside the cluster.

Ok, so we're aligned regarding that.

charts/openshift-etcd-backup/values.yaml
```diff
@@ -0,0 +1,92 @@
...
+  # never (do not expire backups, keep all of them)
+  expiretype: "days"
+  # -- Retention period
+  keepdays: "30"
+  # -- Count retention if expiretype set to count
+  keepcount: "10"
+  # -- Backup schedule
+  schedule: "0 0 * * *"
+
+persistence:
```
tongpu, line 21: @vmaillot @paraenggu Maybe we should "limit" the options to either using provisioning via a storage class or directly specifying the way it is expected to be specified in [`volume`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#volume-v1-core), e.g.:

```yaml
nfs:
  path: /backup
  server: example.com
```

or

```yaml
hostPath:
  path: /mnt/etcd-backup
```

charts/openshift-etcd-backup/templates/pv.yaml
vmaillot: > > As long as we're able to specify the [required GIDs](https://docs.openshift.com/container-platform/4.7/storage/persistent_storage/persistent-storage-nfs.html#storage-persistent-storage-nfs-group-ids_persistent-storage-nfs) that's fine,
> 
> If this is required to make it work, we need to add the ability to configure the Security Context to the chart, because I believe it's not yet implemented. This is not related to volume mounting at all.

It is implemented [here](https://github.com/adfinis-sygroup/helm-charts/blob/feat/add-openshift-etcd-backup-chart/charts/openshift-etcd-backup/values.yaml#L65) but should I just then added the the [supplementalGroups](https://docs.openshift.com/container-platform/4.7/storage/persistent_storage/persistent-storage-nfs.html#storage-persistent-storage-nfs-group-ids_persistent-storage-nfs) setup?

charts/openshift-etcd-backup/templates/cronjob.yaml
```diff
@@ -0,0 +1,73 @@
...
+          {{- end}}
+          {{- with .Values.affinity }}
+          affinity:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          hostNetwork: true
+          hostPID: true
+          serviceAccountName: {{ include "openshift-etcd-backup.serviceAccountName" . }}
+          restartPolicy: Never
+          dnsPolicy: ClusterFirst
```
tongpu, line 50: This is the default according to the Kubernetes API documentation: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#podspec-v1-core

I would suggest to remove it.

```suggestion
```

charts/openshift-etcd-backup/templates/cronjob.yaml
```diff
@@ -0,0 +1,73 @@
...
+              type: Directory
+            name: host
+          {{- if or .Values.persistence.nfs.enabled .Values.persistence.hostPath.enabled .Values.persistence.provisioning.enabled }}
+          - name: volume-backup
+            {{- if .Values.persistence.nfs.enabled }}
+            nfs:
+              server: {{ .Values.persistence.nfs.server | quote }}
+              path: {{ .Values.persistence.nfs.path | quote }}
+            {{- end }}
+            {{- if .Values.persistence.provisioning.enabled }}
```
tongpu, line 64: Technically only one of the persistent storage options should be allowed at a time, so using `else if` here would be preferred.

```suggestion
            {{- else if .Values.persistence.provisioning.enabled }}
```
I would love to have a template added to `_helpers.tpl` that would ensure that only one of them is enabled, but Go doesn't provide XOR, so it would be pretty hard to read for nearly not benefit.

charts/openshift-etcd-backup/templates/cronjob.yaml
```diff
@@ -0,0 +1,73 @@
...
+            persistentVolumeClaim:
+              claimName: {{ include "openshift-etcd-backup.fullname" . }}
+            {{- end }}
+            {{- if .Values.persistence.hostPath.enabled }}
```
tongpu, line 68: Please verify that indentation is still correct after this change.

```suggestion
            {{- else if .Values.persistence.hostPath.enabled }}
```

charts/openshift-etcd-backup/templates/pvc.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.persistence.provisioning.enabled }}
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ include "openshift-etcd-backup.fullname" . }}
+  labels:
+    {{- include "openshift-etcd-backup.labels" . | nindent 4 }}
+spec:
+  {{- if .Values.persistence.provisioning.enabled }}
```
tongpu, line 9: The wrong value has been used in the if.

```suggestion
  {{- if .Values.persistence.provisioning.storageClass }}
```

charts/openshift-etcd-backup/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v2
+name: openshift-etcd-backup
+description: Chart for openshift-etcd-backup solution
+type: application
+version: 0.1.0
+appVersion: 0.1.0
```
tongpu, line 6: If we're going to release 1.0.0 of adfinis-sygroup/openshift-etcd-backup we should probably use that version here.

tongpu (changes requested):
Please fix the typo in the `if` in `templates/pvc.yaml` and remove the empty `templates/NOTES.txt` or add some content to it.

Other then that just some minor details where I would recommend some changes, but those are not critical.

charts/openshift-etcd-backup/templates/pv.yaml
tongpu: To be honest I'm not exactly sure what's going to happen when we run privileged and with uid 0 anyway. If this turns out to be a problem we can add it later.

charts/openshift-etcd-backup/templates/cronjob.yaml
```diff
@@ -0,0 +1,73 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ include "openshift-etcd-backup.fullname" . }}
+spec:
+  schedule: {{ .Values.backup.schedule | quote }}
+  jobTemplate:
+    spec:
+      backoffLimit: 0
```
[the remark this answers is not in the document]
vmaillot: right, i'm not sure about the idea behind that tbh.
cyru might know more, but it's risky and we could remove that i think!

eyenx:
oops, I'm sorry wrong window... I updated the branch :/

eyenx:
I should go back to school and learn some proper `git`

vmaillot (commented):
the chart has been tested with `hostPath` and `provisioning` and works fine!

charts/openshift-etcd-backup/templates/pvc.yaml
eyenx: No. I guess what @tongpu means here is, if the StorageClass is set in provisioning, then add it to the pvc.yaml, but if not, keep it empty and it will take the default storageclass.

vmaillot: but that's what i do right?!

eyenx: not exactly, you're setting a storageClassName even if no storageClass is defined in `provisioning.storageClass`

eyenx: If I set provisoning.enabled to true, but keep the storageClass empty, it would resutlt in adding `storageClassName: ""` would that work? I'm not sure.

vmaillot: yes it does work, if the `storageClass` is empty, it uses the default one

vmaillot: but i will test and clean that out as well, to be more consistent

eyenx: Just tried it out...

The first one is created with `StorageClassName: ""` while the second one has no `StorageClassName` at all in `spec:`

```
NAME   STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   AGE
tmp    Pending                                                     27s
tmp2   Pending                                      default        2s
```

charts/openshift-etcd-backup/templates/cronjob.yaml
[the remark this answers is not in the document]
eni23: since im getting notifyed and annoyed by this tread i can answer as well: yes this is intended. if the script fails, it will fail the 2nd time as well normally. since it is writing and deleting stuff on an volume, just retrying can be dangerous. if it is failing, someone has to check whats gone wrong, that was the idea behind.

btw: wont answer again, unsubscribed for now.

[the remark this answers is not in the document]
eyenx: Thanks Cyru! 

> since im getting notifyed and annoyed by this tread i can answer as well

That's on you, fix your notification settings.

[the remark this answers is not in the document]
vmaillot: haha sorry for that @eni23 and thanks for the explanation!

eyenx (approved):

charts/openshift-etcd-backup/templates/cronjob.yaml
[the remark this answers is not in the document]
tongpu: @eni23 That was my expectation why it is configured that way. Thanks for confirming. :)

charts/openshift-etcd-backup/templates/pvc.yaml
tongpu: When not specifying the `storageClassName` parameter you either get not storage at all, because no default Storage Class has been defined, or the storage would be created using the default Storage Class: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1
The behavior of using `storageClassName: ""` is different and should never provision storage using the default Storage Class.

charts/openshift-etcd-backup/templates/cronjob.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: {{ include "openshift-etcd-backup.fullname" . }}
+spec:
+  schedule: {{ .Values.backup.schedule | quote }}
+  jobTemplate:
+    spec:
```
tongpu, line 8: Since we have feedback from Cyrill that the `backoffLimit` has been added for a specific reason we should add it again, but this time with a comment explaining why.
```suggestion
    spec:
      # Retrying this job is not considered safe, because of that we fail the
      # job immediately if the first execution fails.
      backoffLimit: 0
```

tongpu (changes requested):
The famous one last change. :grimacing:

Since we now have the feedback from Cyrill that there's a reason for the `backoffLimit` we should incorporate it.

Thanks for incorporating all the changes and keeping up with all the discussions that this chart created!

eyenx (approved):

charts/openshift-etcd-backup/templates/cronjob.yaml
eyenx: https://github.com/adfinis-sygroup/helm-charts/pull/287/commits/87c794a9293c1a97f684ea5f3506090d17582a1b

eyenx (approved):
Friday Merge is best merge

charts/openshift-etcd-backup/Chart.yaml
eyenx: Releasing is now done automatically. Because we already had tag v1.0.0 we are now on 1.1.0 after merging the workflow stuff (Feat)

eyenx:
:fireworks:

PR #295, opened 2021-07-06
feat(openshift-etcd-backup): add podSecurityContext support

vmaillot:
# Description

Add support of `podSecurityContext` for additionnalGroups support as described [here](https://docs.openshift.com/container-platform/4.7/storage/persistent_storage/persistent-storage-nfs.html#storage-persistent-storage-nfs-group-ids_persistent-storage-nfs).

# Issues

# Checklist


charts/openshift-etcd-backup/templates/cronjob.yaml
```diff
@@ -11,6 +11,8 @@ spec:
       backoffLimit: 0
       template:
         spec:
+          securityContext:
+            supplementalGroups: {{ .Values.securityContext.supplementalGroups }}
```
hairmare, line 15:
```suggestion
          securityContext:
            {{- toYaml .Values.podSecurityContext | nindent 12 }}
```

charts/openshift-etcd-backup/values.yaml
```diff
@@ -65,6 +65,8 @@ securityContext:
   privileged: true
   # -- Set user ID
   runAsUser: 0
+  # -- Define supplemental groups
+  supplementalGroups: [""]
```
hairmare, line 69:
```suggestion

podSecurityContext:
  # -- Define supplemental groups
  supplementalGroups: [""]
```
do we need an empty string as default here or would an empty array (or even an empty `podSecurityContext`) make more sense?

hairmare (changes requested):

charts/openshift-etcd-backup/values.yaml
vmaillot: you're right, an ampty `podSecurityContext` would fit best!

charts/openshift-etcd-backup/templates/cronjob.yaml
```diff
@@ -11,6 +11,8 @@ spec:
       backoffLimit: 0
       template:
         spec:
+          securityContext:
+             {{- toYaml .Values.podSecurityContext | nindent 12 }}
```
tongpu, line 15: One space too much here:
```suggestion
            {{- toYaml .Values.podSecurityContext | nindent 12 }}
```

charts/openshift-etcd-backup/values.yaml
```diff
@@ -66,6 +66,10 @@ securityContext:
   # -- Set user ID
   runAsUser: 0
 
+podSecurityContext:
```
tongpu, line 69:
```suggestion
# -- Configure SecurityContext of the pod started by the job
podSecurityContext: {}
```

tongpu (changes requested):

charts/openshift-etcd-backup/templates/cronjob.yaml
vmaillot: wow, nice catch, thanks!

hairmare (approved):

tongpu (approved):

PR #319, opened 2021-08-11
Fix for openshift-etcd-backup cronjob

eni23:
# Description

For every CronJob, the CronJob Controller checks how many schedules it missed in the duration from its last scheduled time until now. If there are more than 100 missed schedules, then it does not start the job and logs the error. If the value is set, the cronjob will survive a cluster downtime.

See: https://access.redhat.com/solutions/3667021 and kubernetes/kubernetes#42649 for details

# Checklist


hairmare (approved):
LGTM if tests don't fail

vmaillot (approved):

PR #322, opened 2021-08-17
feat(timed): add static file volume

winged:
The timed backend (especially the admin section) needs some static files
to work.  Given Django best-practices, it is not recommended for the
static files to be served by Django itself, but rather by a dedicated
web server.

This creates a volume for the static files, which can then be shared with
the frontend web server, which already provides the application assets.

The static files are generated by the startup script of the backend
container, so they will be freshly written any time the conainer comes
up. We don't need to worry about them being stale etc.

# Checklist

### DIFF
--- a/charts/infra-apps/examples/thanos.yaml
+++ b/charts/infra-apps/examples/thanos.yaml
@@ -1,7 +1,7 @@
 # Deploy Thanos
 #
-# This takes care of deploying Thanos' store gateway, query frontend, and receiver. 
-# Usually the prometheus-operator from kube-prometheus-stack will deploy 
+# This takes care of deploying Thanos' store gateway, query frontend, and receiver.
+# Usually the prometheus-operator from kube-prometheus-stack will deploy
 # ThanosRulers and sidecars to complete the stack.
 
 thanos:
--- a/.github/workflows/pr-sizing.yaml
+++ b/.github/workflows/pr-sizing.yaml
@@ -1,7 +1,7 @@
 ## Reference: https://github.com/pascalgn/size-label-action
 ---
 name: 'PR Size'
-on: 
+on:
   pull_request_target:
     types: [opened, synchronize, reopened]
 jobs:
--- a/charts/timed/Chart.yaml
+++ b/charts/timed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.7.2
+version: 0.8.0
 appVersion: v1.3.0
 keywords:
   - timed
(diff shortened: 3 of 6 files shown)


charts/infra-apps/examples/thanos.yaml
```diff
@@ -1,7 +1,7 @@
 # Deploy Thanos
 #
-# This takes care of deploying Thanos' store gateway, query frontend, and receiver. 
-# Usually the prometheus-operator from kube-prometheus-stack will deploy 
+# This takes care of deploying Thanos' store gateway, query frontend, and receiver.
+# Usually the prometheus-operator from kube-prometheus-stack will deploy
```
hairmare, line 4:
```suggestion
# This takes care of deploying Thanos' store gateway, query frontend, and receiver. 
# Usually the prometheus-operator from kube-prometheus-stack will deploy 
```

hairmare, line 4: somehow pre-commit did some whitespace changes on files you didn't indend to touch...
```suggestion
# This takes care of deploying Thanos' store gateway, query frontend, and receiver. 
# Usually the prometheus-operator from kube-prometheus-stack will deploy 
```
I'm not sure why we've been seeing this on this repo lately, but am planning to  migrate to pre-commit.ci instead of the current setup which has exhibited these issues over the last couple of weeks.

(the files are indeed wrong but non-timed changes should not be in a timed pr)

hairmare:
> Given Django best-practices, it is not recommended for the static files to be served by Django itself, but rather by a dedicated web server.

I think these best-practices do not apply to this case here. The app is a DRF app and the django admin backend only has a tiny amount of users which wouldn't put much strain on hosting a couple of statics from the backend. Hosting statics from the backend won't work if you have thousands of users and need to have perfect caching, both of these situations do not apply to our case.

IMO the additional complexity this PR adds to the deployment (an additional volume and some logic in more containers) is way more costly than just using `django.contrib.staticfiles` to directly host the files.

winged:
> > Given Django best-practices, it is not recommended for the static files to be served by Django itself, but rather by a dedicated web server.
> 
> I think these best-practices do not apply to this case here. The app is a DRF app and the django admin backend only has a tiny amount of users which wouldn't put much strain on hosting a couple of statics from the backend. Hosting statics from the backend won't work if you have thousands of users and need to have perfect caching, both of these situations do not apply to our case.
> 
> IMO the additional complexity this PR adds to the deployment (an additional volume and some logic in more containers) is way more costly than just using `django.contrib.staticfiles` to directly host the files.

TBH I share the sentiment, and given the very limited user base of the Django admin, it's surely the better way forward. I'm just wondering what'd be the right way if we did indeed have some static files that need to come from the backend. Some people seem to use S3 for static file storage, which would be even more overkill here..

Anyway, thanks for the feedback, I'll see that we can serve the files form Django instead

hairmare:
> I'm just wondering what'd be the right way if we did indeed have some static files that need to come from the backend. Some people seem to use S3 for static file storage, which would be even more overkill here... 

Using a blob storage that is internet-accessible (like S3 or Azure blob) can be a nice architecture pattern but in that case you would usually host a complete SPA using that. So putting just statics onto a blob volume is more or less the classic CDN pattern which IMO isn't per-se bad.

PR #326, opened 2021-08-31
feat(azure-apps): Add azure-apps for azure related components

isantospardo:
# Description

- Creates an *azure-apps* subchart where to store specific Azure applications
- Adds secrets-store-csi-driver-provider-azure for vault

# Checklist


charts/azure-apps/examples/secrets-store-csi-driver-provider-azure.yaml
```diff
@@ -0,0 +1,8 @@
+secretsStoreCsiDriverAzure:
+  enabled: true
+  project: infra-secrets-store-csi-azure
+  values:
+    # Install RBAC roles and bindings required for K8S Secrets syncing if true
+    syncSecret:
```
eyenx, line 6: This stuff should be under `secrets-store-csi-driver` as shown [here](https://github.com/Azure/secrets-store-csi-driver-provider-azure/blob/master/charts/csi-secrets-store-provider-azure/values.yaml#L98)

charts/azure-apps/values.yaml
```diff
@@ -0,0 +1,19 @@
+# secretsStoreCsiDriverAzure -- [secrets-store-csi-driver-provider-azure](https://github.com/Azure/secrets-store-csi-driver-provider-azure) ([example](./examples/secrets-store-csi-driver-provider-azure.yaml))
+# @default -- -
+secretsStoreCsiDriverAzure:
+  # secretsStoreCsiDriverAzure.enabled -- Enable secrets-store-csi-driver-provider-azure
+  enabled: false
+  name: secrets-store-csi-driver-provider-azure
```
eyenx, line 6: perhaps we should rather go for a naming concept like vault did:

https://github.com/hashicorp/vault-csi-provider

so: `azure-key-vault-csi-provider`  or `azure-kv-csi-provider` 

What are your thoughts? Also on the Namespace and Project name.

eyenx (changes requested):

charts/azure-apps/values.yaml
isantospardo: Yes, I think the same, also, the shortest the better, so we can proceed with `azure-kv-csi-provider`. I will update, thanks for the suggestion.

charts/azure-apps/examples/secrets-store-csi-driver-provider-azure.yaml
isantospardo: Missed it, thanks!

eyenx:
I would also consider renaming the app name and all the files in examples if we are going with `azure-kv-csi-provider` probably use `azureKvCsiProvider` instead of the whole `SecretsStoreCsi...`

eyenx (changes requested):
see above

charts/azure-apps/examples/secrets-store-csi-driver-provider-azure.yaml
```diff
@@ -0,0 +1,9 @@
+azureKvCsiProvider:
```
eyenx, line 1: should we also rename this files to match `azure-kv-csi-provider` ?

eyenx (changes requested):

charts/azure-apps/examples/secrets-store-csi-driver-provider-azure.yaml
isantospardo: Yes, why not, to be consistent, thanks for the suggestion

eyenx (approved):

hairmare (commented):
i concur with toni as in the the naming should probably match the repo name more closely

charts/azure-apps/templates/NOTES.txt
```diff
@@ -0,0 +1,4 @@
+The following apps are available:
+{{ if .Values.azureKvCsiProvider.enabled }}
+* azureKvCsiProvider
```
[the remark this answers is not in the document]
isantospardo: I changed the name to `secrets-store-csi-driver-provider-azure`, thanks for the suggestion.

PR #371, opened 2021-09-15
fix(ci): replace pre-commit action with working script

hairmare:
# Description

The pre-commit action we were using was replaced pre-commit.ci which does not offer all the features needed on our side. This  PR replaces the action with a hack to run pre-commit directly without using the action (which was pretty much a glorified shell wrapper anyway).

With this pre-commit should get executed properly again and it will also complain if it's not being used. This doesn't fix charts that contain error introduced while the action was broken, but it will complain once those charts get touched again.

If/once we merge this i will also take care or cleaning up charts that are in a non-pre-commit-limiting state.

The azure-apps change is in here because i needed something to validate that pre-commit works again and it's changes are small enough for us to not need to care about them. The caasperli changes are in here since we validate all the READMEs on each change, doing that is fast enough compared to most other pre-commit tasks.

# Issues
nope
# Checklist


sinicis (approved):
LGTM

Issue #382, opened 2021-09-28
[openshift-etcd-backup] Upgrade to 1.5.x image


davidkarlsen:
**Is your feature request related to a problem? Please describe.**
How about upgrading to the 1.5 image and support defining the prometheus rule with this chart?

**Describe the solution you'd like**
Add opt-in monitoring of jobs with prometheusrule.

**Describe alternatives you've considered**
N/A

**Affected chart**
openshift-etcd-backup

**Additional context**
https://github.com/adfinis-sygroup/openshift-etcd-backup/pull/16

PR #384, opened 2021-09-29
chore(openshift-etcd-backup): update to latest app version and add PrometheusRule

vmaillot:

# Description

Update to latest application version and add PrometheusRule feature.

# Issues

#382

# Checklist


hairmare (changes requested):

charts/openshift-etcd-backup/templates/prometheusrule.yaml
```diff
@@ -0,0 +1,15 @@
...
+  name: {{ include "openshift-etcd-backup.fullname" . }}-cronjob-monitor
+spec:
+  groups:
+  - name: openshift-etcd-backup
+    rules:
+    - alert: EtcdBackupCronJobStatusFailed
+      expr: |
+        kube_job_status_succeeded{namespace="{{ .Release.Namespace }}"} == 0
+      labels:
+        severity: critical
```
hairmare, line 14: please move the actual definition to values.yaml (like [how the argo-cd helm chart does it](https://github.com/argoproj/argo-helm/blob/e4f8dd58109eefe53c41acd98ab50fd443643b22/charts/argo-cd/values.yaml#L188-L190))

This way tuning the rule or adding more rules is much easier.

eni23: @hairmare i dont like this at all, other charts doing it as @vmaillot did here

tongpu: I'm having mixed feelings about that. I'd rather go the way of `kube-prometheus-stack` and how the handle it with the [`defaultRules`](https://github.com/prometheus-community/helm-charts/blob/2cacc16807caedc6cabf1606db27e0d78c844564/charts/kube-prometheus-stack/values.yaml#L33) parameter.

vmaillot: that looks neat! i'll try with that solution!

hairmare: `kube-prometheus-stack` also supports `additionalPrometheusRulesMap` to add more rules (which is what i was after)

eni23: For me this is perfectly fine. But i think what @hairmare was after is an `additionalRules` for adding more extra rules. But we should keep it simple and not forget what this is, its not like this chart is deploying a full prom stack or argocd as in the examples, it just one single pod cronjob wich runs periodically.

hairmare: yeah. i want to be able to add additional rules so we can validate them for a while before we add them directly in here... one rule that comes to mind would check `kube_cronjob_spec_suspend` to alert us if the job gets suspended

vmaillot: just to be sure, should i add the support of [this](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/additionalPrometheusRules.yaml)?

hairmare: it doesn't need to be as complicated as the upstream kps chart does it, but it sure would be nice to be able to add a simple rule for checking `kube_cronjob_spec_suspend` on the fly where needed.

hairmare (changes requested):
I would like to be able to add additional rules on the fly using `additionalRules`. This will help if we need additional customer specific rules but it can also be used if we want to change the included rule (ie. we can disable the included rule and replace it with a tweaked one if need be).

eni23 (approved):
nice, this LGTM

Issue #369, opened 2021-09-14
batch/v1beta1 CronJob deprecation


hairmare:
**Affected charts**
* back8sup:0.3.3
* timed:0.7.3
* rmd:0.1.13
* openshift-etcd-backup:1.4.3

**Describe the bug**

batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob

**To Reproduce**
1. Make changes to Chart
2. Create PR and check the output of the "Lint and Test Charts / Lint and Install" CI action
3. Scroll through "Run chart-testing (install)"
4. See deprecation warning from above description

**Expected behavior**
Charts should support `batch/v1` CronJobs.

**Logs**
<details>
  <summary>Click to expand!</summary>

```
2021-09-14T11:31:28.9001453Z ------------------------------------------------------------------------------------------------------------------------
2021-09-14T11:31:28.9002366Z  Charts to be processed:
2021-09-14T11:31:28.9003676Z ------------------------------------------------------------------------------------------------------------------------
2021-09-14T11:31:28.9004620Z  timed => (version: "0.7.3", path: "charts/timed")
2021-09-14T11:31:28.9012146Z ------------------------------------------------------------------------------------------------------------------------
2021-09-14T11:31:28.9023893Z 
2021-09-14T11:31:28.9692972Z "adfinis" already exists with the same configuration, skipping
2021-09-14T11:31:29.0521855Z "bitnami" already exists with the same configuration, skipping
2021-09-14T11:31:29.1361758Z "secrets-store-csi-driver" already exists with the same configuration, skipping
2021-09-14T11:31:29.2780045Z Hang tight while we grab the latest from your chart repositories...
2021-09-14T11:31:29.3660161Z ...Successfully got an update from the "secrets-store-csi-driver" chart repository
2021-09-14T11:31:29.4842125Z ...Successfully got an update from the "adfinis" chart repository
2021-09-14T11:31:31.3841672Z ...Successfully got an update from the "bitnami" chart repository
2021-09-14T11:31:31.3845331Z Update Complete. ⎈Happy Helming!⎈
2021-09-14T11:31:34.0881671Z Saving 1 charts
2021-09-14T11:31:34.0951461Z Downloading postgresql from repo https://charts.bitnami.com/bitnami
2021-09-14T11:31:35.9859177Z Deleting outdated charts
[...]

PR #394, opened 2021-10-12
fix(back8sup): CronJob apiVersion based on kubernetesVersion

vmaillot:

# Description

Set apiVersion for CronJobs based on Kubernetes version.

# Issues

#369

# Checklist


sinicis (approved):
LGTM

charts/back8sup/values.yaml
```diff
@@ -2,6 +2,8 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# kubeVersionOverride -- override what version of Kubernetes to render against
+kubeVersionOverride: ~
```
tongpu, line 6: What's the use-case for this variable? If you want to manually set a specific version then `helm template` has the parameter `--kube-version` available for just that.

eyenx: Not sure if argoCD uses --template

vmaillot: i changed the way to check for the `apiVersion` looks better now

sinicis (approved):

PR #439, opened 2021-11-18
fix(back8sup): use correct apiVersion and upgrade to lastest patch version

vmaillot:

# Description

This PR fix `apiVersion` when having this error:

```
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "CronJob" in version "batch/v1"
```

# Issues

# Checklist

### DIFF
--- a/charts/back8sup/templates/cronjob.yaml
+++ b/charts/back8sup/templates/cronjob.yaml
@@ -1,4 +1,4 @@
-{{- if .Capabilities.APIVersions.Has "batch/v1" }}
+{{- if .Capabilities.APIVersions.Has "batch/v1/CronJob" }}
 apiVersion: batch/v1
 {{- else }}
 apiVersion: batch/v1beta1
--- a/charts/back8sup/Chart.yaml
+++ b/charts/back8sup/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: back8sup
 description: Deploy back8sup to a Kubernetes Cluster
 type: application
-version: 0.4.1
-appVersion: v0.7.5
+version: 0.4.2
+appVersion: v0.7.6
 home: https://github.com/adfinis-sygroup/back8sup
 sources:
   - https://github.com/adfinis-sygroup/back8sup
--- a/charts/back8sup/README.md
+++ b/charts/back8sup/README.md
@@ -1,6 +1,6 @@
 # back8sup
 
-![Version: 0.4.1](https://img.shields.io/badge/Version-0.4.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.7.5](https://img.shields.io/badge/AppVersion-v0.7.5-informational?style=flat-square)
+![Version: 0.4.2](https://img.shields.io/badge/Version-0.4.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v0.7.6](https://img.shields.io/badge/AppVersion-v0.7.6-informational?style=flat-square)
 
 Deploy back8sup to a Kubernetes Cluster
 


charts/back8sup/templates/cronjob.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Capabilities.APIVersions.Has "batch/v1" }}
+{{- if .Capabilities.APIVersions.Has "batch/v1/CronJob" }}
```
tongpu, line 1: But wasn't this exactly the thing that we realized doesn't work with Argo CD?

vmaillot: well we thought so, but i had this issue for a customer, and just tested it, it works like this... because `batch/v1` exists there but but for the `CronJob` kind.

hairmare: I remember looking this up in the Argo CD source code and coming to the conclusion that it doesn't support `<type>/<version>/<kind>`. For sure something here is inconsistent and needs more debugging given that this PR is basically a revert of #408.

vmaillot: that's right, and it's totally a revert, which fixes our issue at the customer side

vmaillot: oh and we use latest patch version of the application itself!

hairmare: yeah, but will the revert not break whatever was the reason for #408?

vmaillot: well, i tested it with @isantospardo on the customer side, it works like this and we're running argocd v2.1.2 there!

isantospardo: Yep,  I can confirm the fix on the customer side. Does anyone know any other customer running this? so we can test it there, from my side, I don't know any customer using it.

vmaillot: so, approve/not approve?

vmaillot: here you go:

```
~/s/helm/helm-charts/charts/back8sup | fix/back8sup-apiversion | $ h template --api-versions=batch/v1/CronJob . | grep '# Source: back8sup/templates/cronjob.yaml' -A2
# Source: back8sup/templates/cronjob.yaml
apiVersion: batch/v1
kind: CronJob
~/s/helm/helm-charts/charts/back8sup | fix/back8sup-apiversion | $ h template --api-versions=batch/v1 . | grep '# Source: back8sup/templates/cronjob.yaml' -A2       
# Source: back8sup/templates/cronjob.yaml
apiVersion: batch/v1beta1
kind: CronJob
~/s/helm/helm-charts/charts/back8sup | fix/back8sup-apiversion | $ h template --api-versions=batch/v1beta1/CronJob . | grep '# Source: back8sup/templates/cronjob.yaml' -A2
# Source: back8sup/templates/cronjob.yaml
apiVersion: batch/v1beta1
kind: CronJob
~/s/helm/helm-charts/charts/back8sup | fix/back8sup-apiversion | $ h template --api-versions=batch/v1beta1 . | grep '# Source: back8sup/templates/cronjob.yaml' -A2       
# Source: back8sup/templates/cronjob.yaml
apiVersion: batch/v1beta1
kind: CronJob
```

vmaillot: totallly, it means that just `batch/v1` could not be having the `CronJob` in that `apiVersion`.
does it making sense?

eyenx (approved):

charts/back8sup/templates/cronjob.yaml
eyenx: We might have seen how the template command is executed by the repo-server wrong. Le'ts see! LGTM

vmaillot: cheerz, i'll roll that out to the customer!

PR #445, opened 2021-11-29
feat(security-apps): Add dex-k8s-authenticator

isantospardo:
# Description

Due to gangway deprecation
FFI: https://github.com/heptiolabs/gangway#vmware-has-ended-active-development-of-this-project-this-repository-will-no-longer-be-updated
# Issues

Gangway deprecation

# Checklist

### DIFF
--- a/charts/security-apps/values.yaml
+++ b/charts/security-apps/values.yaml
@@ -18,6 +18,29 @@ dex:
   # @default -- [upstream values](https://github.com/dexidp/helm-charts/tree/master/charts/dex/values.yaml)
   values: {}
 
+# dexK8sAuthenticator -- [dex-k8s-authenticator](https://github.com/mintel/dex-k8s-authenticator) ([example](./examples/dex-k8s-authenticator.yaml))
+# @default -- -
+dexK8sAuthenticator:
+  # dexK8sAuthenticator.enabled -- Enable dex-k8s-authenticator
+  enabled: false
+  name: dex-k8s-authenticator
+  destination:
+    # dexK8sAuthenticator.destination.namespace -- Namespace
+    namespace: "infra-dex-k8s-authenticator"
+  # dexK8sAuthenticator.repoURL -- Repo URL
+  # @default -- [repo](https://github.com/mintel/dex-k8s-authenticator.git)
+  repoURL: "https://github.com/mintel/dex-k8s-authenticator.git"
+  # dexK8sAuthenticator.repoPath -- Repo Path
+  repoPath: "charts/dex-k8s-authenticator"
+  # dexK8sAuthenticator.chart -- Chart
+  chart: "dexK8sAuthenticator"
+  # dexK8sAuthenticator.targetRevision -- [dex-k8s-authenticator Helm chart](https://github.com/mintel/dex-k8s-authenticator/tree/master/charts/dex-k8s-authenticator/) version
+  targetRevision: "v1.4.*"
+  # dexK8sAuthenticator.values -- Helm values
+  # @default -- [upstream values](https://github.com/mintel/dex-k8s-authenticator/tree/master/charts/dex-k8s-authenticator/values.yaml)
+  values: {}
+
+# This repo is DEPRECATED, use dexK8sAuthenticator instead
 # gangway -- [gangway](https://github.com/heptiolabs/gangway/) ([example](./examples/gangway.yaml))
 # @default -- -
 gangway:
--- a/charts/security-apps/values.yaml
+++ b/charts/security-apps/values.yaml
@@ -27,6 +27,9 @@ dexK8sAuthenticator:
   destination:
     # dexK8sAuthenticator.destination.namespace -- Namespace
     namespace: "infra-dex-k8s-authenticator"
+  # As of December 2021, there is no upstream helm repo where we could pull
+  # for this we need to reference directly the GitHub url
+  # FFI: https://github.com/mintel/dex-k8s-authenticator/issues/185
   # dexK8sAuthenticator.repoURL -- Repo URL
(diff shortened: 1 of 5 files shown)


charts/security-apps/values.yaml
```diff
@@ -18,6 +18,29 @@ dex:
...
+  chart: "dexK8sAuthenticator"
+  # dexK8sAuthenticator.targetRevision -- [dex-k8s-authenticator Helm chart](https://github.com/mintel/dex-k8s-authenticator/tree/master/charts/dex-k8s-authenticator/) version
+  targetRevision: "v1.4.*"
+  # dexK8sAuthenticator.values -- Helm values
+  # @default -- [upstream values](https://github.com/mintel/dex-k8s-authenticator/tree/master/charts/dex-k8s-authenticator/values.yaml)
+  values: {}
+
+# This repo is DEPRECATED, use dexK8sAuthenticator instead
 # gangway -- [gangway](https://github.com/heptiolabs/gangway/) ([example](./examples/gangway.yaml))
 # @default -- -
```
hairmare, line 45: please mark up the change so it shows up in the README
```suggestion
# gangway -- [gangway](https://github.com/heptiolabs/gangway/) is DEPRECATED, use dexK8sAuthenticator instead
# @default -- DEPRECATED
```

charts/security-apps/values.yaml
```diff
@@ -18,8 +18,30 @@ dex:
...
+dexK8sAuthenticator:
+  # dexK8sAuthenticator.enabled -- Enable dex-k8s-authenticator
+  enabled: false
+  name: dex-k8s-authenticator
+  destination:
+    # dexK8sAuthenticator.destination.namespace -- Namespace
+    namespace: "infra-dex-k8s-authenticator"
+  # dexK8sAuthenticator.repoURL -- Repo URL
+  # @default -- [repo](https://github.com/mintel/dex-k8s-authenticator.git)
+  repoURL: "https://github.com/mintel/dex-k8s-authenticator.git"
```
vmaillot, line 32: i found a valid repo in https://artifacthub.io/packages/helm/sagikazarmark/dex-k8s-authenticator which would be `https://charts.sagikazarmark.dev`

isantospardo: Yeah, but that is pointing to a user's repo.. dont think we should use any user's repo anywhere, this could impact the infrastucture in case the user wants to delete the image or even create security issues. I think we should stick to the main repo or open some issues and provide the solution for this, not sure what the others think about it..

vmaillot: that makes sense! thanks
but could be the same with the github repo right? :)

isantospardo: Yep, but at least is the offical repo, so we somehow get more guarantee, but again, this is only my opinion, doesn't mean this is the way to go :)

vmaillot: right, looks good to me then :)

vmaillot (approved):

charts/security-apps/values.yaml
```diff
@@ -18,8 +18,30 @@ dex:
...
+  # dexK8sAuthenticator.repoPath -- Repo Path
+  repoPath: "charts/dex-k8s-authenticator"
+  # dexK8sAuthenticator.chart -- Chart
+  chart: "dexK8sAuthenticator"
+  # dexK8sAuthenticator.targetRevision -- [dex-k8s-authenticator Helm chart](https://github.com/mintel/dex-k8s-authenticator/tree/master/charts/dex-k8s-authenticator/) version
+  targetRevision: "v1.4.0"
```
hairmare, line 38: we should really look for a real repo so we can wildcard some:
```suggestion
  targetRevision: "1.4.*"
```

otherwise this feels very hacky

charts/security-apps/values.yaml
hairmare, line 32: official chart repo seems to be here:
```suggestion
  repoURL: "https://mintel.github.io/helm-charts/"
```

hairmare (dismissed):
please switch to the charts official helm repo as per my comments above

charts/security-apps/values.yaml
hairmare: imo this needs a comment that  explains the rationale behind it being the 1 (one) -apps chart that uses a github directly and thus can't wildcard semver versions.

the comment has to explain that there is no repo for the thing and it must also include some links to upstream issues discussing this.

isantospardo:
@hairmare thanks for checking, yeah, I don't see any indexes there, so should I proceed directly pointing to the version, right?

hairmare:
> @hairmare thanks for checking, yeah, I don't see any indexes there, so should I proceed directly pointing to the version, right?

yes, but add a comment to the code that explains the situation in detail and include a bunch of links to upstreams discussions regarding helm charts

if you could also check if upstream if there are any recent updates for upstreams helm charting efforts that would be awesome

charts/security-apps/values.yaml
isantospardo: let me know if you think I should add more coments in there

charts/security-apps/values.yaml
isantospardo: Upstream helm chart repo never finished
FFI: https://github.com/mintel/dex-k8s-authenticator/issues/185

charts/security-apps/values.yaml
isantospardo: Upstream helm chart repo never finished
FFI: https://github.com/mintel/dex-k8s-authenticator/issues/185

hairmare: thanks for reaching out to upstream and adding comments!

hairmare (approved):

PR #475, opened 2021-12-13
feat(argo-apps): migrate argocd-notifications to argo-apps

hairmare:
# Description

⚠️ unless you used argocd notifications (and you probably didn't) this is a no-op change

Create a new `argo-apps` and move Argo CD Notifications to it (from `infra-apps`). 

AFAICT no one has used it in infra apps so far, i also plan on adding some more argo things (like argocd-applicationsets and rollouts) to the new `argo-apps` down the line.

# Issues

# Checklist


charts/argo-apps/ci/default-values.yaml
```diff
@@ -0,0 +1,4 @@
+argocdNotifications:
+  enabled: true
+  values:
+    TODO: add values in values.yaml
```
tongpu, line 4: What's the idea behind this `TODO` line?

tongpu (commented):
This will create a new release for `infra-apps`, because of that I would suggest to explicitly mention at the top of the PR that this is an "empty" release of `infra-apps`.

Have we thought about how we could maintain changelogs for our charts in some way, because that would make it easier to track changes in GitHub Releases and maybe also ArtifactHub?

charts/argo-apps/ci/default-values.yaml
hairmare: tbh, it's mostly cargo culted from elsewhere

hairmare:
> suggest to explicitly mention at the top of the PR

@tongu I updated the PR and will also add a note in the commit during merge

tongpu (approved):

Issue #84, opened 2020-09-09
[*-apps] We need usage docs for argoconfig and *-apps charts


hairmare:
**Affected chart**
All (potential) `*-apps` charts

**Describe the bug**
The auto-generated documentation for the charts is currently lacking. The `examples/` dirs in the charts have proven to not contain enough details for end users.

**Expected behavior**
There should be nice end-user quality docs for `*-apps` charts and the used corresponding library charts.

**Additional context**
The `*-apps` charts aim at being as generic as possible so documenting them in the top-level `docs/` probably makes most sense.

**Tasks**
* [ ] initialize `*-apps` docs
* [ ] explain how to use argoconfig
* [ ] explain app-of-apps concept
* [ ] explain initial deployment & boostrap (esp. regarding `infra-apps` and the `(env-)main-argo` app)
* [ ] document how to use `$app.name` to override namespace based names

eyenx:
@tongpu suggested documenting values like `syncPolicy` inside the `-apps/values.yaml` but I'm against this, as this values come from the `argoconfig` library chart.

How do we want to make sure people know, they can set these values? Extend argoconfig's documentation is enough?

hairmare:
At some we might want to add some high-level docs to the `docs/` folder. Those might describe how to use the app-of-apps things based on argoconfig.

hairmare:
@vmaillot there seem to be several tasks that were not addressed

PR #477, opened 2021-12-13
docs(app-of-apps): argoconfigs template

hairmare:
# Description

Add a section on the `argoconfig` lilbrary chart.

# Issues

- https://github.com/adfinis-sygroup/helm-charts/issues/84

# Checklist

* [x] ~~I updated the version in Chart.yaml~~
* [x] ~~I updated applicable README.md files using  `pre-commit run`~~
* [x] ~~If I updated a dependency tool, or app, this PR contains a short summary of the changes I'm pulling~~
* [x] ~~I am ready to test changes after they are applied and released~~


tongpu (commented):
I'll have to read the second part again tomorrow, but here are some minor recommendations from my side regarding how to lead over to code snippets. I would use `:`, but I'm actually not sure if the English language might be different from German in this regard.

docs/argocd-app-of-apps.md
```diff
@@ -26,6 +26,45 @@ we consider valuable on most clusters. As with all the charts, `infra-apps` can
...
+# ... common.metadata
+spec:
+  project: "default"
+  source: {}
+  destination:
+    server: "https://kubernetes.default.svc"
+  syncPolicy: {}
+```
+
+Our app-of-apps charts invoke the function in their individual `Application` templates:
```
tongpu, line 48: After a nights sleep I'm now able to phrase what was bothering me yesterday. What I'm missing here is the information how to configure an application contained in an app-of-apps chart, e.g. how does the structure look like in `values.yaml` and what needs to be configured where.

This could possibly be a topic on top of the `argoconfig` section, because it would allow the reader to read on from a usage documentation to the exact implementation.

Do you understand what I'm aiming for?

hairmare: see b89b794df89b51c4c3ac0203ae45aee005aeb64c

docs/argocd-app-of-apps.md
```diff
@@ -26,6 +26,57 @@ we consider valuable on most clusters. As with all the charts, `infra-apps` can
...
+  {{- end }}
+  # ... more overriding, check any app-of-apps chart for an example
+{{- end -}}
+```
+
+The first block in this example renders the `Application`, the second block injects a template used for rendering. The first thing that
+is done in the second block is to assign everything specific to an app to `$app` so it can be referenced in the template without needing
+to write out the full `.Values.example` part over and over.
+
+In the above example the values section would contain a minimal example section as default:
```
tongpu, line 70:
```suggestion
In the above example the `values.yaml` for an app-of-apps chart would contain a minimal section for an `example` application by default:
```

docs/argocd-app-of-apps.md
```diff
@@ -26,6 +26,57 @@ we consider valuable on most clusters. As with all the charts, `infra-apps` can
...
+
+```yaml
+example:
+  enabled: false
+  name: example
+  # ... more values, check any app-of-apps chart for an example
+  value: {}
```
tongpu, line 77:
```suggestion
  values: {}
```

tongpu (commented):
I think we're still not at the point where we're able to really explain the concept in layman's terms, but I'm not sure if this is even possible.

hairmare:
> I think we're still not at the point where we're able to really explain the concept in layman's terms, but I'm not sure if this is even possible.

This really only focuses on the library chart bits, #84 has some more tasks that should get us closer to explaning the big picture concept.

tongpu (approved):

PR #520, opened 2022-01-17
feat(infra-apps): argocd chart 3.30

hairmare:
# Description

Argo CD chart 3.30.*. Adds google managed certs on GKE.

# Issues

* https://github.com/argoproj/argo-helm/pull/1088

# Checklist


tongpu (approved):
Not even sure if we need to integrate this update of the Helm chart, but better to have it then not.

hairmare:
> Not even sure if we need to integrate this update of the Helm chart, but better to have it then not.

i'm pretty sure we don't need the feature anywhere so it's more of a freebie update and not a real adding of features.

PR #541, opened 2022-01-26
fix(timed): fix ingress template for apiVersion v1

trowik:
# Description

Ingress template wasn't updated correctly for the api version v1.

# Issues

# Checklist

### DIFF
--- a/charts/timed/templates/ingress.yaml
+++ b/charts/timed/templates/ingress.yaml
@@ -17,25 +17,35 @@ spec:
         paths:
           - path: /admin/
             backend:
-              serviceName: {{ $.Values.backend.service.name }}
-              servicePort: {{ $.Values.backend.service.externalPort }}
+              service:
+                name: {{ $.Values.backend.service.name }}
+                port:
+                  name: http
           - path: /static/
             backend:
-              serviceName: {{ $.Values.backend.service.name }}
-              servicePort: {{ $.Values.backend.service.externalPort }}
+              service:
+                name: {{ $.Values.backend.service.name }}
+                port:
+                  name: http
           - path: /api/
             backend:
-              serviceName: {{ $.Values.backend.service.name }}
-              servicePort: {{ $.Values.backend.service.externalPort }}
+              service:
+                name: {{ $.Values.backend.service.name }}
+                port:
+                  name: http
           - path: /oidc/
             backend:
-              serviceName: {{ $.Values.backend.service.name }}
-              servicePort: {{ $.Values.backend.service.externalPort }}
+              service:
+                name: {{ $.Values.backend.service.name }}
+                port:
+                  name: http
           # frontend service as catch all backend
           - path: /
             backend:
-              serviceName: {{ $.Values.frontend.service.name }}
-              servicePort: {{ $.Values.frontend.service.externalPort }}
+              service:
+                name: {{ $.Values.frontend.service.name }}
+                port:
+                  name: http
     {{- end -}}
   {{- if .Values.ingress.tls }}
   tls:
--- a/charts/timed/Chart.yaml
+++ b/charts/timed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.8.1
+version: 0.8.2
 appVersion: v1.4.5
 keywords:
   - timed
--- a/charts/timed/README.md
+++ b/charts/timed/README.md
@@ -1,6 +1,6 @@
 # timed
 
-![Version: 0.8.1](https://img.shields.io/badge/Version-0.8.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v1.4.5](https://img.shields.io/badge/AppVersion-v1.4.5-informational?style=flat-square)
+![Version: 0.8.2](https://img.shields.io/badge/Version-0.8.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v1.4.5](https://img.shields.io/badge/AppVersion-v1.4.5-informational?style=flat-square)
 
 Chart for Timed application
 
--- a/charts/timed/templates/service-backend.yaml
+++ b/charts/timed/templates/service-backend.yaml
@@ -11,7 +11,7 @@ spec:
     - port: {{ .Values.backend.service.externalPort }}
       targetPort: http
       protocol: TCP
-      name: {{ .Values.backend.service.name }}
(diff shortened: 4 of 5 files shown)


hairmare (approved):
in theory caasperli has an example of how to patch the ingress in a way that it supports both versions, but this is fine for timed as well

LGTM

charts/timed/templates/ingress.yaml
```diff
@@ -17,25 +17,35 @@ spec:
         paths:
           - path: /admin/
             backend:
-              serviceName: {{ $.Values.backend.service.name }}
-              servicePort: {{ $.Values.backend.service.externalPort }}
+              service:
+                name: {{ $.Values.backend.service.name }}
+                port:
+                  number: {{ $.Values.backend.service.externalPort }}
```
tongpu, line 23: I'd recommend to change to named ports instead of using the port number:
```suggestion
                  name: {{ $.Values.backend.service.name }}
```

charts/timed/templates/ingress.yaml
```diff
@@ -17,25 +17,35 @@ spec:
...
           - path: /static/
             backend:
-              serviceName: {{ $.Values.backend.service.name }}
-              servicePort: {{ $.Values.backend.service.externalPort }}
+              service:
+                name: {{ $.Values.backend.service.name }}
+                port:
+                  number: {{ $.Values.backend.service.externalPort }}
```
tongpu, line 29: I'd recommend to change to named ports instead of using the port number:
```suggestion
                  name: {{ $.Values.backend.service.name }}
```

charts/timed/templates/ingress.yaml
```diff
@@ -17,25 +17,35 @@ spec:
...
           - path: /api/
             backend:
-              serviceName: {{ $.Values.backend.service.name }}
-              servicePort: {{ $.Values.backend.service.externalPort }}
+              service:
+                name: {{ $.Values.backend.service.name }}
+                port:
+                  number: {{ $.Values.backend.service.externalPort }}
```
tongpu, line 35: I'd recommend to change to named ports instead of using the port number:
```suggestion
                  name: {{ $.Values.backend.service.name }}
```

charts/timed/templates/ingress.yaml
```diff
@@ -17,25 +17,35 @@ spec:
...
           - path: /oidc/
             backend:
-              serviceName: {{ $.Values.backend.service.name }}
-              servicePort: {{ $.Values.backend.service.externalPort }}
+              service:
+                name: {{ $.Values.backend.service.name }}
+                port:
+                  number: {{ $.Values.backend.service.externalPort }}
```
tongpu, line 41: I'd recommend to change to named ports instead of using the port number:
```suggestion
                  name: {{ $.Values.backend.service.name }}
```

charts/timed/templates/ingress.yaml
```diff
@@ -17,25 +17,35 @@ spec:
...
           # frontend service as catch all backend
           - path: /
             backend:
-              serviceName: {{ $.Values.frontend.service.name }}
-              servicePort: {{ $.Values.frontend.service.externalPort }}
+              service:
+                name: {{ $.Values.frontend.service.name }}
+                port:
+                  number: {{ $.Values.frontend.service.externalPort }}
```
tongpu, line 48: I'd recommend to change to named ports instead of using the port number:
```suggestion
                  name: {{ $.Values.frontend.service.name }}
```

vmaillot (approved):

tongpu (changes requested):
Changing to named ports instead of port numbers probably would avoid some fuckups in the future.

charts/timed/templates/ingress.yaml
trowik: changed to named ports

charts/timed/templates/ingress.yaml
trowik: changed to named ports

charts/timed/templates/ingress.yaml
trowik: changed to named ports

charts/timed/templates/ingress.yaml
trowik: changed to named ports

charts/timed/templates/ingress.yaml
trowik: changed to named ports

vmaillot:
lgtm!

vmaillot (approved):

charts/timed/templates/ingress.yaml
hairmare: `$.Values.backend.service.name` contains the name of the service and not the port. Meethinks this should be hardcoded `http` instead.

it's hardcoded in the deployment as well:

https://github.com/adfinis-sygroup/helm-charts/blob/448b4536de7f8eb54ada4abc8bcb922fdec29cbb/charts/timed/templates/deployment-backend.yaml#L47

hairmare (changes requested):
As per my inline comment, If you use named ports then the name needs to be `http` and not `.Values.service.name`.

tongpu:
> As per [my inline comment](https://github.com/adfinis-sygroup/helm-charts/pull/541#discussion_r792798924), If you use named ports then the name needs to be `http` and not `.Values.service.name`.

`http` is the name for the port in the container. In ingress you have to configure the name of the port on the service, which is currently `{{ .Values.backend.service.name }}`: https://github.com/adfinis-sygroup/helm-charts/blob/448b4536de7f8eb54ada4abc8bcb922fdec29cbb/charts/timed/templates/service-backend.yaml#L14

But this could be changed to be `http`, which I didn't propose, because I deemed it out of scope of this change.

trowik:
> > As per [my inline comment](https://github.com/adfinis-sygroup/helm-charts/pull/541#discussion_r792798924), If you use named ports then the name needs to be `http` and not `.Values.service.name`.
> 
> `http` is the name for the port in the container. In ingress you have to configure the name of the port on the service, which is currently `{{ .Values.backend.service.name }}`:
> 
> https://github.com/adfinis-sygroup/helm-charts/blob/448b4536de7f8eb54ada4abc8bcb922fdec29cbb/charts/timed/templates/service-backend.yaml#L14
> 
> But this could be changed to be `http`, which I didn't propose, because I deemed it out of scope of this change.


So what's the best approach? Should the name of the port be hardcoded as `http` in service-backend  or in ingress?

tongpu:
> So what's the best approach? Should the name of the port be hardcoded as `http` in service-backend or in ingress?

I would hardcode the name for the port in both services to `http` and use `name: http` in the ingress definition. This makes the port naming consistent in all objects (ingress, service, container) using it.

hairmare (approved):

PR #546, opened 2022-01-31
feat(tracing-apps): install opentelemetry-collector

hairmare:
# Description

Install the [OpenTelemetry Collector](https://opentelemetry.io/docs/collector/).

# Issues

# Checklist

### DIFF
--- a/charts/tracing-apps/Chart.yaml
+++ b/charts/tracing-apps/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tracing-apps
 description: Argo CD app-of-apps config for tracing applications
 type: application
-version: 0.7.1
+version: 0.8.0
 home: https://github.com/adfinis-sygroup/helm-charts/tree/master/charts/tracing-apps
 sources:
   - https://github.com/adfinis-sygroup/helm-charts
--- a/charts/tracing-apps/README.md
+++ b/charts/tracing-apps/README.md
@@ -1,6 +1,6 @@
 # tracing-apps
 
-![Version: 0.7.1](https://img.shields.io/badge/Version-0.7.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square)
+![Version: 0.8.0](https://img.shields.io/badge/Version-0.8.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square)
 
 Argo CD app-of-apps config for tracing applications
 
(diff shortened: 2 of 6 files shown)


tongpu:
What's the plan for this? Are we already using this anywhere?

hairmare:
> What's the plan for this? Are we already using this anywhere?

We are not using it anywhere yet but in a lot of cases it should be between the app and ie. a jaeger deployment. I have a couple of casees where i expect it too make sense soon (as in i told devs to use the otel-sdk for their language) and am expecting this to see some use then.

upstream describes it as follows:

> The OpenTelemetry Collector offers a vendor-agnostic implementation of how to receive, process and export telemetry data. It removes the need to run, operate, and maintain multiple agents/collectors. This works with improved scalability and supports open-source observability data formats (e.g. Jaeger, Prometheus, Fluent Bit, etc.) sending to one or more open-source or commercial back-ends. The local Collector agent is the default location to which instrumentation libraries export their telemetry data.

For the cases where i see us using the collector it helps with decoupling the actual implementation of the back-end from collection. This means the instrumentation library in the instrumented code can be more lightweight since it does not need to take care of things like backoffs and knowing about multiple backends.

isantospardo (approved):
From the point of view of the deployment, it looks good to me, but probably better to wait for @tongpu approval due to the ongoing discussions.

hairmare:
@isantospardo rerequesting a review was not intended, i was fishing for a way to rerequest one from @tongpu

tongpu (approved):
> > What's the plan for this? Are we already using this anywhere?
> 
> We are not using it anywhere yet but in a lot of cases it should be between the app and ie. a jaeger deployment. I have a couple of casees where i expect it too make sense soon (as in i told devs to use the otel-sdk for their language) and am expecting this to see some use then.

Ok, this sound to me like the fluentd of Telemetry. :D

If this allows the developers to consume the only the OpenTelemetry library and the data can be sent to different tracing backends then you're good to go.

tongpu:
> isantospardo rerequesting a review was not intended, i was fishing for a way to rerequest one from @tongpu

That probably didn't work, because my comment was not part of a review and because of that I didn't consume the review you requested yet.

hairmare:
OpenTelemetry (and the collector) isn't just about tracing, it does full stack instrumentation including logs and metrics if you want to :)

tongpu:
> OpenTelemetry (and the collector) isn't just about tracing, it does full stack instrumentation including logs and metrics if you want to :)

Just don't call it observability or I feel obliged to summon mipsytipsy onto you!

hairmare:
:partying_face: o11y change merged :partying_face:

PR #548, opened 2022-02-02
feat(chart-testing): introduce pluto

eyenx:
# Description

Introduce Pluto for warning about depreacted / removed API Versions

# Issues

n/a

# Checklist

---


hack/pluto.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# wrapper to call pluto
+
+set -e
+
+source hack/sh/rc.sh
+source hack/sh/_functions.sh
```
hairmare, line 8: 🤔

hack/sh/deps/pluto.sh
```diff
@@ -0,0 +1,19 @@
+#/bin/bash
+
+set -e
+
+source hack/sh/_functions.sh
```
hairmare, line 5: which function are you using? it looks  like none so why include _functions?

hairmare:
do we need something in `docs/` wrt introducing pluto?

hack/sh/deps/pluto.sh
hairmare: ah, `err`

hack/pluto.sh
hairmare: which function are you using? it looks like none so why include _functions?

hack/pluto.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# wrapper to call pluto
+
+set -e
+
+source hack/sh/rc.sh
+source hack/sh/deps/pluto.sh
+
+helm template $1 | pluto detect -
```
hairmare, line 10: ok, so i realised why there was no `err`.

we should do something like this to handle errors:
```suggestion
source hack/sh/_functions.sh
source hack/sh/deps/pluto.sh

chart_dir=$1

[[ -d $chart_dir ]] || err "No chart dir given. Usage: hack/pluto.sh chart/name"

helm template $chart_dir | pluto detect -
```

Issue #131, opened 2020-12-14
[rbac-definitions] No RBAC Definition App for RBAC Manager


eyenx:
**Is your feature request related to a problem? Please describe.**
We are deploying rbac-manager with infra-apps here: https://github.com/adfinis-sygroup/helm-charts/blob/master/charts/infra-apps/values.yaml#L83

But ther is no app managing the definitions. We should have an app defining the rbac definitions as follows:

https://github.com/FairwindsOps/rbac-manager/blob/master/examples/rbacdefinition-sa-imagepull.yaml

**Describe the solution you'd like**
Additional rbac-definition app inside infra-apps

**Describe alternatives you've considered**
Managing rbac-definitions directly with rbac-manager app, but there is no way of doing this with upstream rbac-manager chart:
https://github.com/FairwindsOps/charts/blob/master/stable/rbac-manager/values.yaml

**Affected chart**
`infra-apps/rbac-manager`

**Additional context**
--

eyenx:
For the moment managed as RAW YAML

hairmare:
As per the docs the `-apps` charts are just a collection of the (chart_repo, chart_name, chart_version) tuple.

As such, for the proposed solution, there would need to be a chart available to add to `infra-apps` as part of this issue. I didn't look very far, but i don't think such a chart currently exists.

As for the considered alternative, did upstream every decline adding the resources to their chart in a vein similar to other `-stack` style charts?

eyenx:
I created an issue https://github.com/adfinis-sygroup/helm-charts/issues/131

eyenx:
> As for the considered alternative, did upstream every decline adding the resources to their chart in a vein similar to other `-stack` style charts?

Yes, Upstream declined today the implementation. Even referenced cert-manager where we also have a custom Chart for issuers. 

I'll go ahead and create a new chart then.

eyenx:
#581

PR #581, opened 2022-02-26
feat(rbacdefinitions): new chart

eyenx:
# Description

Introduces a new chart for deploying rbacDefinitions via Helm

# Issues

solves #131

# Checklist

* [ ] If I updated a dependency tool, or app, this PR contains a short summary of the changes I'm pulling

---


charts/rbacdefinitions/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
+name: rbacdefinitions
```
hairmare, line 2: can we have dash in the name please?
```suggestion
name: rbac-definitions
```

charts/rbacdefinitions/templates/rbacdefinitions.yaml
```diff
@@ -0,0 +1,13 @@
+{{- range .Values.rbacDefinitions }}
+---
+# raw rbacDefinition: {{ . | toJson }}
```
hairmare, line 3:
```suggestion
```

(i forgot to remove this debug line in cert-manager-issuers but it's wrong there as  well)

charts/rbacdefinitions/Chart.yaml
eyenx: yeah was bugging me too during the whole creation. glad you think the same. I am changing this later.

charts/rbacdefinitions/templates/rbacdefinitions.yaml
eyenx: haha I was unsure why it was there, so I just left it. gonna remove it.

charts/rbacdefinitions/Chart.yaml
hairmare: i was even considering `rbac-manager-definitions` but it might get a tad to long with that.

charts/rbac-definitions/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
+name: rbac-definitions
+description: Deploy RBACDefinitions via Helm
```
hairmare, line 3:
```suggestion
description: Deploy rbac-manager RBACDefinitions via Helm
```

charts/rbac-definitions/templates/NOTES.txt
```diff
@@ -0,0 +1,6 @@
+{{- if .Values.rbacDefinitions }}
+Deployed RBACDefinitions:
+{{- range .Values.rbacDefinitions }}
```
hairmare, line 3: wouldn't this need to match what is happening in the definitions: 
```suggestion
{{- range .Values.rbacDefinitions }}
{{- $enabled := printf "%t" .enabled }}
{{- if regexMatch "nil" $enabled }}{{- $enabled = "true" }}{{- end }}
{{- if (eq $enabled "true") }}
```

hairmare (changes requested):
In addition to the feedback below, i think we should also add a test for the `enabled` thing in `charts/rbac-definitions/tests/rbacdefinitions_test.yaml`.

eyenx:
Yeah, it would be explicit for rbac-manager. Rbac-definitions could
confuse some folks you're right. But it would be too long. Perhaps
having a better description in the Chart? 


-- 

Adfinis AG
Antonio Tauro, System Engineer, GPG KeyID: 0x0796132F0077A5F8
Güterstrasse 86 | CH-4053 Basel
Office +41 61 500 31 31 | Direct +41 61 500 31 37
www.adfinis.com

charts/rbac-definitions/templates/NOTES.txt
eyenx: yeah you are right, it should match what is in the template, although I don't have any `enabled` value in the `Values.yaml` we need to fix this also in `cert-issuers` too then.

I'll apply this suggestion from you, and add an additional `test` to check it out.

isantospardo (approved):

charts/rbac-definitions/Chart.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: v2
+name: rbac-definitions
+description: Deploy rbac-manager RBACDefinitions via Helm
+type: application
+version: 0.1.0
+# matches apiVersion: rbacmanager.reactiveops.io/v1beta1
+appVersion: "1beta1"
+keywords:
+  - rbac-manager
+home: https://rbac-manager.docs.fairwinds.com/
```
hairmare, line 10: it might make sense to deeplink the actual docs of the CRD
```suggestion
home: https://rbac-manager.docs.fairwinds.com/rbacdefinitions/
```

hairmare (approved):

PR #582, opened 2022-02-28
docs(infra-apps): document setting ingress as default class

hairmare:
# Description

Add the values needed to configure the ingress that is part of `infra-apps` as default. This should always be the default on cluster where only one ingress exists (which is the case in most deployments).

# Issues

# Checklist

### DIFF
--- a/charts/infra-apps/examples/nginx-ingress-on-aks.yaml
+++ b/charts/infra-apps/examples/nginx-ingress-on-aks.yaml
@@ -1,7 +1,8 @@
 # nginx-ingress on AKS
 #
 # Deploys the nginx-ingress component using and external load balancer
-# like is the default on AKS.
+# like is the default on AKS. Also marks the ingress as the default so
+# it may be uses without explicitly specifying it.
 
 ingressNginx:
   enabled: true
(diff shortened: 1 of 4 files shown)


charts/infra-apps/examples/nginx-ingress-on-aks.yaml
```diff
@@ -1,7 +1,8 @@
 # nginx-ingress on AKS
 #
 # Deploys the nginx-ingress component using and external load balancer
-# like is the default on AKS.
+# like is the default on AKS. Also marks the ingress as the default so
+# it may be uses without explicitly specifying it.
```
tongpu, line 5:
```suggestion
# it may be used without explicitly specifying it.
```

charts/infra-apps/examples/nginx-ingress.yaml
```diff
@@ -1,7 +1,8 @@
 # Deploy nginx-ingress with a customer default page
 #
 # This nxing-ingress example shows how you can specify a custom default
-# backend.
+# backend. Also marks the ingress as the default so
+# it may be uses without explicitly specifying it.
```
tongpu, line 5:
```suggestion
# it may be used without explicitly specifying it.
```

tongpu (commented):
Some typos I came across.

eyenx (approved):

PR #610, opened 2022-03-25
feat(vault-monitoring): add VaultAutopilotNodeHealthy rule

in0rdr:
# Description

Add an alert rule for Raft integrated storage VaultAutopilotNodeHealthy

To be done:
- [x] This is only applicable to integrated storage installations. Not sure about the behavior for other storage backends.
- [x] Where do we need to place a note about this? Probably in in `docs/`?

# Issues

n/a

# Checklist


in0rdr:
> Could we find a solution for different cluster sizes here? As this configuration seems to only apply to 5 node clusters.

https://github.com/adfinis-sygroup/helm-charts/issues/611

eyenx (changes requested):

charts/vault-monitoring/values.yaml
```diff
@@ -50,6 +50,17 @@ prometheusRules:
   # prometheusRules.rules -- set of prometheus alerts to define
   # @default -- list of predefined alerts
   rules:
+    - alert: VaultAutopilotNodeHealthy
+      # Set to 1 if the node_id is deemed healthy by Autopilot, 0 if not
+      # node_id: Unique identifier of a raft peer, same as peer_id
+      # https://www.vaultproject.io/docs/internals/telemetry#integrated-storage-raft-autopilot
+      expr: count by (instance) (vault_autopilot_node_healthy) < 5
```
eyenx, line 57:
```suggestion
      expr: vault_autopilot_healthy < 1
```

As per https://www.vaultproject.io/docs/internals/telemetry#integrated-storage-raft-autopilot we can just use `vault_autopilot_healthy`

eyenx: Fixes #611

in0rdr: Thanks, that would also resolve @pree question regarding the node count. I'll test this on a cluster to observe the behavior..

in0rdr: This works perfectly fine, I tested it. Also, changed the description to make clear that the check still has the same "amount of information" (it checks the health of all the nodes and alerts "if at least one node is considered unhealthy"), even without the explicit threshold.

eyenx (approved):

in0rdr:
Also performed a local test with a Vault using ["file"](https://www.vaultproject.io/docs/configuration/storage/filesystem) storage backend.

The fact that some storage backends don't use Raft is not an issue.

The alert is just never ["active"](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/) and is listed as `VaultAutopilotNodeHealthy (0 active)` which is fine.

eyenx:
@in0rdr can we merge this?

eyenx:
bump @in0rdr

pree:
Does `vault_autopilot_healthy` also report correctly when a node leaves & re-joins? There will be one node unhealthy in Prometheus for a short time when this happens, but a n+1th node which would be healthy.

eyenx:
@in0rdr @pree BUMP please update the pr and merge if possible.

in0rdr:
@eyenx I need more time to properly answer this question. I understand that there are two distinct test cases:

- 1 node unexpectedly looses the connection to the cluster
- 1 node leaves the cluster ([remove-peer](https://www.vaultproject.io/docs/commands/operator/raft#remove-peer), deliberate)

The check proposed in this PR works well for the first case. However, I'm not sure about the behavior on deliberate "node leaves & [re-join](https://www.vaultproject.io/docs/commands/operator/raft#join)" actions.

Could be the case that this check does not even detect such actions (remove-peer, join), but I would need to test this in detail to find out.

charts/vault-monitoring/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: vault-monitoring
 description: monitor your vault server from within Kubernetes' prometheus
 type: application
-version: 0.1.4
-home: https://github.com/adfinis-sygroup/helm-charts/tree/main/charts/vault-monitoring
+version: 0.2.0
+home: https://github.com/adfinis-sygroup/helm-charts/tree/master/charts/vault-monitoring
```
hairmare, line 6: it looks like automerge picked the wrong side of the merge for this line:
```suggestion
home: https://github.com/adfinis-sygroup/helm-charts/tree/main/charts/vault-monitoring
```

eyenx:
Any update on this @in0rdr  ?

in0rdr:
Hey @eyenx and @pree,

Thanks for bearing with me, finally found time to fix my playground and give this a try to answer our question.

Turns out, we cannot use `vault_autopilot_node_healthy` or `vault_autopilot_healthy` to detect nodes that leave the raft peer list (`vault operator raft remove-peer`). In other words, it is not an effective check, if you would like to detect when an operator deliberately reconfigures the raft peer list (e.g., remove 1/3 peers with `remove-peer` command).

The `node_healthy` checks detect, when one of the nodes (in the current peer list) goes down and is no longer able to communicate with the cluster through port `8201`. However, this only works, as long as the node is part of the peer list. To make an example, the following test order does not trigger an alarm:

1. Remove 1/3 peers with `remove-peer` command
2. Shutdown the removed node 

Whereas a simple shutdown (without step 1) does trigger the alarm, the test above will not alert, because the cluster continues operation with 2/3 nodes (the node was removed deliberately).

(Sidenote: As long as the nodes communicate through cluster ports `8201`, raft autopilot and peer list will show the information of the previously joined cluster)

Use [`vault.raft.peers`](https://www.vaultproject.io/docs/internals/telemetry#integrated-storage-raft-autopilot) or similar to detect nodes that left the cluster with `remove-peer` command. But careful here, this again requires an exact threshold (e.g., 3 or 5).

Please note that this behavior does not change, if we enable "Cleanup Dead Servers" with "Min Quorum" and "Dead Server Last Contact Threshold" (set to a very short time like 1m for testing).

I hope the behavior is clear now. Happy to conduct further testing later this week (preferably offline).

hairmare:
>  In other words, it is not an effective check, if you would like to detect when an operator deliberately reconfigures the raft peer list (e.g., remove 1/3 peers with remove-peer command).

this sounds very much like the behavior we would expect, ie. the monitoring should alert if the state differs from what the operator requested and not try to monitor the operator

imo we can always go back and add an additional rule if we need to check for something like a minimum amount of peers (and those depend on the actual deployment so can't be added globally in here w/o a tunable value).

eyenx:
totally agree with @hairmare here

smirta:
> Whereas a simple shutdown (without step 1) does trigger the alarm, the test above will not alert, because the cluster continues operation with 2/3 nodes (the node was removed deliberately).

At least in this scenario something should alert, no? This is lost quorum and according to documentation should result in a sealed cluster. But I see the problem and I'm with @eyenx and @hairmare here. If someone wants to monitor this because he doesn't trust his operators, he should monitor the audit logs. The system monitoring is not intended for something like that.

hairmare (approved):

PR #631, opened 2022-04-07
fix(barman): Adjust slot-creation

ggggut:
# Description

Hello, creating a slot failed on our side because `ERROR: Unknown server 'pg'`.
I found it hardcoded in the cronjob. 
Wouldn't it be better to use the `postgres.host` value for creating the db slot?
Let me know what you think!

Best

# Issues

````bash
barman.server ERROR: Check 'replication slot' failed for server 'xxxxxxx'
	replication slot: FAILED (replication slot 'barman' doesn't exist. Please execute 'barman receive-wal --create-slot xxxxxxx')
	directories: OK
	retention policy settings: OK
````

# Checklist

### DIFF
--- a/charts/barman/Chart.yaml
+++ b/charts/barman/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: barman
 type: application
 description: Chart for Barman PostgreSQL Backup and Recovery Manager
-version: 0.7.0
+version: 0.7.1
 appVersion: "v2.19"
 keywords:
   - barman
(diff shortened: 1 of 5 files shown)


hairmare:
`postgres.host` does feel much better than having a hardcoded `pg` string. It should probably also get surfaced in `values.yaml`. 

For the `barman.backups` thing to work it would probably also need some `if hasKey`/`$.Values` stuff in the template (like what is already there for backupSchedule).

ggggut:
thank you! i will look into it tomorrow

ggggut:
The `barman.backups` key works now. But..

````
It should probably also get surfaced in values.yaml.
````
Can you elaborate?

ggggut:
I also added the possibility to add annotations. This could be useful in a few cases.

hairmare:
> ```
> It should probably also get surfaced in values.yaml.
> ```
> 
> Can you elaborate?

I didn't realise that `.postgres.host` is already defined in `values.yaml` :facepalm:

hairmare (approved):
LGTM

PR #635, opened 2022-04-12
feature(barman): Allow setup of sidecars

ggggut:
# Description

Hello, me again,
what is your opinion on allowing sidecars into the pod?
One interesting usecase would be to access the PVs in an sidecar to do some recovery verification, or automatic restores, etc. 
Also barman supports [Remote recovery](https://docs.pgbarman.org/release/2.19/#remote-recovery).
For this it would also be interesting to have a sidecar to do SSH stuff. 

# Issues

# Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,7 +22,7 @@ jobs:
           git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
 
       - name: Install Helm
-        uses: azure/setup-helm@v1.1
+        uses: azure/setup-helm@v2.1
         with:
           version: v3.8.0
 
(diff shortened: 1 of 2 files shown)


hairmare:
I think we've mostly being adding initContainers to the database that needs restoring to accomplish this, but i'm very much ok with allowing sidecar containers to be created via this chart.

Can you consider calling the value `deployment.extraContainers` rather than `deployment.sidecars`? It feels like this would make it more consistent compared to other helm charts that use some variation on `extra*` for adding stuff.

For SSH, do you also need to expose a port in the sidecar? If yes, then you would probably also want to add a some way to create additional services to make such ports reachable.

hairmare:
> Something like this? :)

I think so, yes :)

tbh i'm not really sure if it's needed, but it certainly won't hurt given it's deactivated by default.

PR #638, opened 2022-04-14
feat(mopsos): add Grafana deployment

hairmare:
# Description

Deploys a Grafana instance complete with pre-configured postgres datasource.

I'm also tweaking some of the default values to clarify that it doesn't make sense to run mopsos with sqlite unless on test.

# Issues

# Checklist

### DIFF
--- a/charts/mopsos/values.yaml
+++ b/charts/mopsos/values.yaml
@@ -163,7 +163,7 @@ grafana:
           # -- `hostname:port` for postgresql server to be used
           url: ~
           database: mopsos
-          user: mopsos
-          password: $POSTRESQL_PASSWORD
+          user: mopsos_ro
+          password: $POSTRESQL_RO_PASSWORD
   # -- Load mopsos secrets into grafana env
   envFromSecret: mopsos-secret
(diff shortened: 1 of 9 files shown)


charts/mopsos/values.yaml
```diff
@@ -127,3 +127,43 @@ config:
...
+      datasources:
+        - name: Mopsos
+          uid: mopsos
+          type: postgres
+          isDefault: true
+          access: proxy
+          # -- `hostname:port` for postgresql server to be used
+          url: ~
+          database: mopsos
+          user: mopsos
```
eyenx, line 166: we want a RO user here, right?

hairmare: yes we do, i'll update everything to match

are you ok with suffix ing ro users with `_ro` (so it would be `mopsos_ro`)?

eyenx: YES

vmaillot (approved):
lgtm! can't wait for the demo!

eyenx (approved):

PR #690, opened 2022-05-25
feat: initial introduction for 'existingHost' variable

luk43:
# Description

Introduce `postgresql.existingHost` variable to define custom postgresql instance.

# Issues

# Checklist

### DIFF
--- a/charts/timed/templates/configmap-backend.yaml
+++ b/charts/timed/templates/configmap-backend.yaml
@@ -53,10 +53,10 @@ data:
 {{- end }}
   DJANGO_OIDC_ADMIN_LOGIN_REDIRECT_URL: {{ .Values.auth.oidc.adminLoginRedirectUrl | quote }}
 
-{{- if .Values.postgresql.enabled }}
+{{- if not .Values.postgresql.existingHost }}
   DJANGO_DATABASE_HOST: {{ include "timed.postgresql.fullname" . | quote }}
 {{- else }}
-  DJANGO_DATABASE_HOST: {{ .Values.postgresql.service.host | quote }}
+  DJANGO_DATABASE_HOST: {{ .Values.postgresql.existingHost | quote }}
 {{- end }}
 {{- if .Values.redmine.enabled }}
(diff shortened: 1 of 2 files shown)


charts/timed/values.yaml
```diff
@@ -231,6 +231,8 @@ ingress:
 postgresql:
   # -- Enable PostgreSQL for persistence
   enabled: true
```
luk43, line 233: We may want to get rid of that, as a DB is used anyway?

hairmare: it's used in Charts.yaml to activated/deactivate the whole postgresql subchart, i don't think we can get rid of it

charts/timed/templates/configmap-backend.yaml
```diff
@@ -53,10 +53,10 @@ data:
 {{- end }}
   DJANGO_OIDC_ADMIN_LOGIN_REDIRECT_URL: {{ .Values.auth.oidc.adminLoginRedirectUrl | quote }}
 
-{{- if .Values.postgresql.enabled }}
+{{- if not .Values.postgresql.existingHost }}
```
hairmare: i'd suggest something like this

```
{{- if .Values.postgresql.enabled }}
  DJANGO_DATABASE_HOST: {{ include "timed.postgresql.fullname" . | quote }}
{{- else if .Values.postgresql.existingHost }}
  DJANGO_DATABASE_HOST: {{ .Values.postgresql.existingHost | quote }}
{{- else }}
  DJANGO_DATABASE_HOST: {{ .Values.postgresql.service.host | quote }}
{{- end }}
```

An ExternalAddress would probably make this a bit less confusing overall.

vmaillot (approved):

eyenx:
needs rebase.

eyenx (changes requested):
rebase

vmaillot:
> rebase

nice catch @eyenx

hairmare (approved):
LGTM once rebased

PR #707, opened 2022-06-15
overhaul existing postgresql instance function

luk43:
# Description

It seperates the config of the postgreSQL-bitnami helm-chart and an existing database connection.
Reason for this is that it can get confusing considering the other values for the bitnami-chart which then aren't active etc.

It also removes the static postgresql dependency.

# Issues

# Checklist


charts/timed/Chart.yaml
```diff
@@ -15,8 +15,3 @@ maintainers:
   - name: adfinis-sygroup
     email: support@adfinis.com
     url: https://adfinis.com
-dependencies:
-  - name: postgresql
-    version: ~9.1.2
-    repository: https://charts.bitnami.com/bitnami
-    condition: postgresql.enabled
```
hairmare, line 22: you will probably still need the dependecy just so there is a database on ci, otherwise chart-testing might fail the install bits

luk43:
Fails with "No chart changes detected.", I reverted stuff. Can we somehow trick it?

eyenx:
> Fails with "No chart changes detected.", I reverted stuff. Can we somehow trick it?

The problem is more the dependency which cannot be built:

```
Save error occurred:  could not find : no matching version
Error: could not find : no matching version
------------------------------------------------------------------------------------------------------------------------
Error: Error linting charts: Error building dependencies for chart 'timed => (version: "0.10.2", path: "charts/timed")': Error waiting for process: exit status 1
```

charts/timed/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.9.1
+version: 0.10.6
```
hairmare, line 5: you only need to bump the version once:
```suggestion
version: 0.10.0
```

hairmare (commented):
i have but one small nitpick about version bumping:

charts/timed/templates/cronjob-backend.yaml
```diff
@@ -32,7 +32,7 @@ spec:
                   valueFrom:
                     secretKeyRef:
                       name: {{ include "timed.postgresql.fullname" $ }}
-                      key: postgresql-password
+                      key: postgres-password
```
eyenx, line 35: isn't this the Postgres Admin PW as default?

https://github.com/bitnami/charts/blob/b1d2449d52bab5b498ea3b56837c6ecd63274fab/bitnami/postgresql/values.yaml#L149

hairmare: yes, but it's only used on ci

hairmare (approved):

charts/timed/templates/cronjob-backend.yaml
eyenx: it's in the `templates` so no, it's not only used on CI. If we would use the chart with the postgres dep, it would reference the Admin PW by default. 

I think this should be `password` instead of `postgres-password`

eyenx (dismissed):
See my comments above

luk43:
> See my comments above

we fixed the build, what do you mean?

eyenx:
> > See my comments above
> 
> we fixed the build, what do you mean?

https://github.com/adfinis-sygroup/helm-charts/pull/707#discussion_r904621139

charts/timed/templates/cronjob-backend.yaml
hairmare: >  If we would use the chart with the postgres dep

This PR is about not using the postgres dep, the only reason the dep is still here is for CI.

eyenx: This PR is about adding support for external postgresql Database. Not making the postgres dep always the admin password

eyenx: Just don't hardcode the usage of admin-password in the templates and I'm happy. If someone wants to deploy timed with this change and use the postgres dep, it would use the admin-password by default and that is not good. We can make it parameterized and add it to values and use ci/default-values.yaml to overwrite it to use the admin-password as key

luk43: Think I changed it everywhere, but somewhere its expecting something different (in testing):
``` psycopg2.OperationalError: FATAL:  password authentication failed for user "postgres" ```

Switching back to WIP, until I figured this out.

File: charts/timed/values.yaml [hunk 4]
```diff
@@ -229,20 +241,15 @@ ingress:
...
-  # postgresqlPassword: s3cr3t
-  image:
-    # Use Postgresql 12
-    # -- PostgreSQL image version to use
-    tag: 12.3.0
+  auth:
+    # -- PostgreSQL database name
+    database: timed
+    # -- PostgreSQL user name
+    username: postgres
```
hairmare, line 250: most likely the initially created user needs renaming as well if we use password instead of postgres-password:
```suggestion
    username: timed
```

luk43: took some time to understand the testing-setup. it may be that the "migrate.py" task might fail because of downgrade of privileges. It could also be that the "migrate.py" already wants to create a user named "timed" (that would be an easy fix).
lets see. hopeful it will just work :D

hairmare: woohoo :) it werked :tada:

hairmare (approved):

PR #708, opened 2022-06-18
Some changes and modifications

mojtabash78:
# Description

- Barman connection to postgres fixed via `scopeName`
- `barman receive-wal` is now done via `barman cron`
- `post_backup_retry_script` and `pre_recovery_retry_script` added to global configuration
- `configmap-entrypoint.yaml` created and `scopeName` added to it
- `additionalENVs` added for deployment
- `values.yaml` file updated
- `README.md` file updated
- `.helmignore` file added 

# Issues

# Checklist


charts/barman/templates/configmap-barman-backups.yaml
```diff
@@ -6,8 +6,8 @@ metadata:
     {{- include "barman.labels" . | nindent 4 }}
 data:
   {{- range .Values.barman.backups }}
-  {{ .postgresql.host }}.conf: |
-      [{{ .postgresql.host }}]
+  {{ .scopeName }}.conf: |
+      [{{ .scopeName }}]
```
tongpu, line 10: What is the benefit of using `scopeName` instead of the hostname?

charts/barman/templates/configmap-barman.yaml
```diff
@@ -19,15 +19,23 @@ data:
...
       compression = {{ .Values.barman.compression }}
       configuration_files_directory = /etc/barman/barman.d
       last_backup_maximum_age = {{ .Values.barman.lastBackupMaximumAge }}
-      minimum_redundancy = 1
+      minimum_redundancy = 0
       ;network_compression = true
       retention_policy = {{ .Values.barman.retentionPolicy }}
       ; retention_policy_mode = auto
-      ;reuse_backup = link
+      ; reuse_backup = link
```
tongpu, line 38:
```suggestion
      ;reuse_backup = link
```

charts/barman/templates/configmap-barman.yaml
```diff
@@ -19,15 +19,23 @@ data:
...
+      {{- if hasKey .Values.barman "preRecoveryRetryScript" }}
+      pre_recovery_retry_script = {{ .Values.barman.preRecoveryRetryScript }}
+      {{- end }}
+
       ; barman_lock_directory = %(barman_home)s
       compression = {{ .Values.barman.compression }}
       configuration_files_directory = /etc/barman/barman.d
       last_backup_maximum_age = {{ .Values.barman.lastBackupMaximumAge }}
-      minimum_redundancy = 1
+      minimum_redundancy = 0
```
tongpu, line 34: Is there a specific reason why you would want to change this to default to 0?
```suggestion
      minimum_redundancy = 1
```

tongpu (changes requested):
Hi @mojtabash78 

Thank you for your contribution. Could you please have a look at the changes that I've proposed and quickly elaborate on the benefits of using an entrypoint script instead of `cron`.

charts/barman/templates/configmap-barman-backups.yaml
mojtabash78: Hi @tongpu 
When I used `.postgresql.host` in my configmaps, I got an error that Barman can't find `pg` which is the `server_name` in Barman commands. Also, Kubernetes domain names for postgresql in my cluster are too long, so it's easier to work with `pg`. Therefore I replaced `.postgresql.host` by `scopeName` in `configmap-barman-backups` and `configmap-barman-cron`.

mojtabash78:
> Hi @mojtabash78
> 
> Thank you for your contribution. Could you please have a look at the changes that I've proposed and quickly elaborate on the benefits of using an entrypoint script instead of `cron`.

Hi @tongpu 
I am using ubc image for Barman: https://github.com/ubc/barman-docker
I suppose you use this image as well.
This image has a [entrypoint.sh](https://github.com/ubc/barman-docker/blob/master/entrypoint.sh) script. In this script, `cron schedules` and `Barman configurations` are generated, but we already generated them in configmaps and this script tries to overwrite them in configmaps which results in error because files mounted by configmaps are read-only. I also set `scopeName` in `configmap-entrypoint`, so Barman commands for replication-status and creating replication slot will be done successfully.

charts/barman/templates/configmap-barman-backups.yaml
hairmare: I think the current state of things was introduced with #631 and it broke quite a few use-cases.

mojtabash78:
> > > Thank you for your contribution. Could you please have a look at the changes that I've proposed and quickly elaborate on the benefits of using an entrypoint script instead of `cron`.
> > 
> > 
> > Hi @tongpu I am using ubc image for Barman: https://github.com/ubc/barman-docker I suppose you use this image as well. This image has a [entrypoint.sh](https://github.com/ubc/barman-docker/blob/master/entrypoint.sh) script. In this script, `cron schedules` and `Barman configurations` are generated, but we already generated them in configmaps and this script tries to overwrite them in configmaps which results in error because files mounted by configmaps are read-only. I also set `scopeName` in `configmap-entrypoint`, so Barman commands for replication-status and creating replication slot will be done successfully.
> 
> Personally i'm not a big fan of overriding the entrypoint with one generated into a configmap, but i see that it is most likely the straight forward way to make the chart be usable for a large range of requirements.
> 
> if i recall correctly the original argument against using a kubernetes-native CronJob was something about the CronJob and barman's internal ssh service being tightly coupled. We investigated using a K8s CronJob a while back and came to the conclusion that too much re-architecting would be needed to make it work.

Hi @hairmare 
Thanks for your descriptions.
I see that you agree that `entrypoint.sh` is getting overridden with one generated in configmap and this indeed prevents Barman to perform the backup procedure. The script is implemented in the `ubc` image (which is used in `values.yaml`) and gets overridden anyway. So how do you suggest to use the chart other than making `entrypoint.sh` a configmap?

Best

charts/barman/templates/configmap-barman-backups.yaml
mojtabash78: Hi @hairmare 
I checked this PR. As I said, when I used `.postgresql.host`, it caused an error implying that Barman can't find server `pg`. I'm not aware that how other friends use `.postgresql.host` and perform backup since Barman is looking for `pg` and I didn't find anything to change this. So I let Barman use `pg` and put it in my configmaps. Sounds right?

hairmare:
> So how do you suggest to use the chart other than making `entrypoint.sh` a configmap?

The only idea i have so far is landing the needed changes directly in the upstream ubc container. That would de-facto mean rewriting to script to support a wide range of args so we can configure it to our liking at runtime. Looking at the current entrypoint, I don't think that would be a reasonable proposal for the ubc image.

Because of that, i think that generating the entrypoint into a ConfigMap and then mounting it is a reasonable approach. This is not something i'd usually recommend, but for this use-case it looks ok.

hairmare:
@tongpu i think this lgtm, do we need any additional feedback from our team to get this merged? I'll raise it in tomorrows stand-up to be sure but i assume we can probably land it soon.

tongpu:
> > @tongpu i think this lgtm, do we need any additional feedback from our team to get this merged? I'll raise it in tomorrows stand-up to be sure but i assume we can probably land it soon.
> 
> Hi @hairmare any updates on this PR ?

I haven't found time last week to look into it again, but I'll provide some feedback until tomorrow.

tongpu (approved):
@mojtabash78 Thanks for your contribution. You probably need to rebase your branch before we can merge it.

hairmare:
Thank you very much for your contribution!

PR #740, opened 2022-06-29
feat(misc-apps): upgrade kube-downscaler from 20.5.0 to 22.2.0

hairmare:
# Description

This updates **kube-downscaler** all the way from 20.5.0 to 22..2.0. It also gets rid of the legacy stable chart repository.

BREAKING CHANGE you will need to adapt any kube-downscaler approjects to use `https://charts.deliveryhero.io` instead of `https://charts.helm.sh/incubator` when applying this change.

The various updates contain a bunch of features. If you want to look at them all then the best way to do so is in [kube-downscaler commit log](https://codeberg.org/hjacobs/kube-downscaler/commits/branch/main). Basically the release fixes a bunch of small problems and adds support for more resources:

* CronJobs
* KEDA ScaledObjects
* Argo Rollouts

# Issues

# Checklist


eyenx (approved):

eyenx:
Hmm by default there is a DEFAULT_UPTIME in the chart.

https://github.com/deliveryhero/helm-charts/blob/master/stable/kube-downscaler/values.yaml#L50

We should be careful with this. We don't overwrite it by default on some of our deploymentns (example our test cluster)

PR #743, opened 2022-07-08
chore(caasperli)!: remove EOL Kubernetes version

eyenx:
# Description

This will remove compatibilty for EOL Kubernetes Versions (<1.21)

Not sure how to bump the version. Should we go for `1.0.0` ? 

# Issues

n/a 

# Checklist

### DIFF
--- a/charts/caasperli/templates/ingress.yaml
+++ b/charts/caasperli/templates/ingress.yaml
@@ -1,7 +1,18 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "caasperli.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
+{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
+  {{- if not (hasKey .Values.ingress.annotations "kubernetes.io/ingress.class") }}
+  {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" .Values.ingress.className}}
+  {{- end }}
+{{- end }}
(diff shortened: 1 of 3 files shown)


charts/caasperli/templates/ingress.yaml
```diff
@@ -23,9 +12,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  {{- if and .Values.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
```
hairmare, line 26: shouldnt this be `if .Values.ingress.className` to keep it optional?

right now this would render the className as an empty string in all cases

charts/caasperli/templates/ingress.yaml
```diff
@@ -43,19 +30,12 @@ spec:
         paths:
           {{- range .paths }}
           - path: {{ .path }}
-            {{- if and .pathType (semverCompare ">=1.18-0" $.Capabilities.KubeVersion.GitVersion) }}
```
hairmare, line 46: as with `ingress.className` we might only want to remove the second bit of the `and`, i'd have to check if pathType is indeed optional in k8s tho

hairmare:
cheap way out of this would probably be to update caasperli with the latest version of what `helm create` does. AFAICT it was updated to do something sane that should support all apiVersion versions.

eyenx:
> cheap way out of this would probably be to update caasperli with the latest version of what `helm create` does. AFAICT it was updated to do something sane that should support all apiVersion versions.

hmm the helm create creates exactly the templates/ingress.yaml we already have.. 

Closing this.

PR #746, opened 2022-07-14
chore(timed): update timed image tags

trowik:
# Description

Update timed image tags

# Issues

# Checklist


charts/timed/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.9.1
+version: 0.9.2
```
tongpu, line 5: I would possibly bump the minor version of the Helm chart when you're updating the major version of the application being deployed by it.
```suggestion
version: 0.10.0
```

trowik: So it should be 0.11.0 then, right? The previous chart version doesn't show correctly as it should be 0.10.0

hairmare (approved):

PR #777, opened 2022-08-30
feat(infra-apps): add oauth2-proxy helm chart

Nerkho:
# Description

Add oauth2-proxy helm chart to infra-apps

# Checklist

* [ x ] This PR contains a description of the changes I'm making
* [ x ] I updated the version in Chart.yaml
* [ x ] I updated applicable README.md files using  `pre-commit run`
* [ x ] I documented any high-level concepts I'm introducing in `docs/`

### DIFF
--- a/charts/infra-apps/values.yaml
+++ b/charts/infra-apps/values.yaml
@@ -269,3 +269,25 @@ mimir:
   # -- Helm values
   # @default -- [upstream values](https://github.com/grafana/mimir/tree/main/operations/helm/charts/mimir-distributed/values.yaml)
   values: {}
+
+# -- [oauth2-proxy] (https://github.com/oauth2-proxy/oauth2-proxy)
+# @default -- [example](./examples/.yaml)
+oauth2-proxy:
+  # -- Enable oauth2-proxy
+  enabled: false
+  name: oauth2-proxy
+  # -- Annotations for oauth2-proxy app
+  annotations: {}
+  desitnation:
+    # -- Namespace
+    namespace: infra-oauth2-proxy
+  # -- Repo URL
+  # @default -- [repo] (https://https://oauth2-proxy.github.io/manifests)
+  repoURL: https://oauth2-proxy.github.io/manifests
+  # -- Chart
+  chart: oauth2-proxy
+  # -- [oauth2-proxy Helm chart] (https://github.com/oauth2-proxy/manifests/tree/main/helm/oauth2-proxy)
+  targetRevision: 6.2.*
+  # -- Helm values
+  # @default -- [upstream values](https://github.com/oauth2-proxy/manifests/blob/main/helm/oauth2-proxy/values.yaml)
+  values: {}
--- a/charts/infra-apps/Chart.yaml
+++ b/charts/infra-apps/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: infra-apps
 description: Argo CD app-of-apps config for infrastructure components
 type: application
-version: 0.111.0
+version: 0.112.0
 home: https://github.com/adfinis/helm-charts/tree/main/charts/infra-apps
 sources:
   - https://github.com/adfinis/helm-charts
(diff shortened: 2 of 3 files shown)


charts/infra-apps/values.yaml
```diff
@@ -269,3 +269,25 @@ mimir:
   # -- Helm values
   # @default -- [upstream values](https://github.com/grafana/mimir/tree/main/operations/helm/charts/mimir-distributed/values.yaml)
   values: {}
+
+# -- [oauth2-proxy] (https://github.com/oauth2-proxy/oauth2-proxy)
+# @default -- [example](./examples/.yaml)
```
hairmare, line 274: can you add a simple example and fix this link?

charts/infra-apps/values.yaml
```diff
@@ -269,3 +269,25 @@ mimir:
...
+oauth2-proxy:
+  # -- Enable oauth2-proxy
+  enabled: false
+  name: oauth2-proxy
+  # -- Annotations for oauth2-proxy app
+  annotations: {}
+  desitnation:
```
hairmare, line 281:
```suggestion
  destination:
```

hairmare (changes requested):
i'm not 100% sure this should really be in infra-apps, infra-apps are things we consider essential for each cluster and oauth2-proxy is more of a component that you need to combine with other bits for it to make sense.

to some degree this is also why my alertmanager deployment is a sidecar and not a full on proper helm based deployment, i'm not sure how much sense that makes tho.

weberval:
@hairmare The thing is, inserting the oauth2-proxy as a sidecar is not the cleanest/prettiest solution. 
Do you have another idea how we could use oath2-proxy as a “not sidecar”?
One could also argument, that secure login to alertmanager should be a key component.

charts/infra-apps/values.yaml
Nerkho: I added an example and fixed the link

charts/infra-apps/values.yaml
Nerkho: fixed the typo here.

hairmare:
> The thing is, inserting the oauth2-proxy as a sidecar is not the cleanest/prettiest solution.

it is indeed quite hacky, that is true.

when i implemented it that way i was expecting to also do something similar for prometheus itself. i noticed that to support both i would have to be able to deploy multiple instances of oauth2-proxy. adding it to infra-apps didn't solve this for me.

i do think that auth is something that alertmanager should support out of the box. either with oauth2-proxy or on it's own. i haven't found time to track down any related issues in alertmanager and start fixing this at the source tho.

so, basically, i'm ok with merging this here, but we need to keep in mind that we'll only be able to deploy and use oauth2-proxy once with this and i'm not sure if once is enough tbh (the other apps in infra-apps have much more of a singleton character as in you only need them once on a typical (simple) cluster).

Nerkho:
> but we need to keep in mind that we'll only be able to deploy and use oauth2-proxy once with this and i'm not sure if once is enough tbh

Working on Kutt last Friday and today, I think we will also need an oauth2-proxy there as well (I need to verify if SSO is a requirement). Making a separate thing, might be the way forward.

hairmare:
> Can we merge this one?

Let's move it to security-apps instead of infra-apps and merge it like that.

Nerkho:
> > Can we merge this one?
> 
> Let's move it to security-apps instead of infra-apps and merge it like that.

Done ! I moved it to security-apps.

eyenx:
We do not have any `ci/default-values.yaml` here?

Nerkho:
> We do not have any `ci/default-values.yaml` here?

Yep missed it. It's added now.

eyenx (approved):

eyenx:
security-apps works for me.

hairmare (approved):

eyenx:
@Nerkho you can merge this now

PR #808, opened 2022-09-20
feat(infra-apps): Update kured to 3.0.1

gianklug:
# Description
Updated the Kured chart to the next major version.
This will probably break most of the installations:
>  The Helm chart labels have been realigned to conform with the [standard labels](https://helm.sh/docs/chart_best_practices/labels/#standard-labels) in the current Helm Chart Best Practices guide, so this upgrade will fail unless the DaemonSet is deleted and recreated. The only way that Helm supports delete and recreate is by uninstalling, so please uninstall the Kured Helm chart before installing again with v3.x.

# Issues
https://github.com/adfinis/helm-charts/issues/793

# Checklist

### DIFF
--- a/charts/infra-apps/Chart.yaml
+++ b/charts/infra-apps/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: infra-apps
 description: Argo CD app-of-apps config for infrastructure components
 type: application
-version: 0.113.0
+version: 0.114.0
 home: https://github.com/adfinis/helm-charts/tree/main/charts/infra-apps
 sources:
   - https://github.com/adfinis/helm-charts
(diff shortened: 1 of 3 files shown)


eyenx:
Can you link us to the source of the changelong?

gianklug:
> Can you link us to the source of the changelog?

It's mentioned in the [upgrade notes](https://github.com/weaveworks/kured/tree/main/charts/kured)

eyenx:
this is also interesting for us 

> A "ServiceMonitor" needs a "release" label to be discovered by the Prometheus-Operator with the default configuration of kube-prometheus-stack and this chart (in the prior v2.x chart) already sets a release label hardcoded. This is changed by applying the best-practise labels in the chart v3.x. Now the user can decide which release label-value should be used.
> With this update, it's more readily possible to make use of the Kured chart with kube-prometheus-stack's default ServiceMonitor selector configuration.

eyenx (approved):
It's not only important to see the changes of the chart here, but also the changes on the image / application version itself.

We are bumping now from `kured:1.10.1` to `kured: 1.10.2` which is just a patchfix release. Nothing is gonna break according to https://github.com/weaveworks/kured/releases/tag/1.10.2 except the helm labels stuff mentioned already. We will need to sync/force/recreate the DS when syncing, but it's not a big thing. 

This is why this LGTM

PR #826, opened 2022-10-06
feat(timed-backend): add probes for django hurricane deployment

fugal-dy:


hairmare (approved):
much nice, very startup probe, such dedicated port for probes

:100: lgtm

PR #905, opened 2022-12-09
feat: [barman] add init containers

kfirfer:
Added initContainers capability
For example waiting for postgresql instance(s)

### DIFF
--- a/charts/barman/Chart.yaml
+++ b/charts/barman/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: barman
 type: application
 description: Chart for Barman PostgreSQL Backup and Recovery Manager
-version: 0.9.2
+version: 0.10.0
 appVersion: "v2.19"
 keywords:
   - barman
@@ -17,3 +17,12 @@ maintainers:
   - name: adfinis
     email: support@adfinis.com
     url: https://adfinis.com
+annotations:
+  artifacthub.io/changes: |
+    - kind: added
+      description: "added init containers for barman deployment"
+      links:
+        - name: GitHub PR
+          url: https://github.com/adfinis/helm-charts/pull/905
+
+
--- a/charts/barman/Chart.yaml
+++ b/charts/barman/Chart.yaml
@@ -24,5 +24,3 @@ annotations:
       links:
         - name: GitHub PR
           url: https://github.com/adfinis/helm-charts/pull/905
-
-
--- a/charts/barman/Chart.yaml
+++ b/charts/barman/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: barman
 type: application
 description: Chart for Barman PostgreSQL Backup and Recovery Manager
-version: 0.9.1
+version: 0.9.2
 appVersion: "v2.19"
 keywords:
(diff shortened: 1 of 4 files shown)


charts/barman/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: barman
 type: application
 description: Chart for Barman PostgreSQL Backup and Recovery Manager
-version: 0.9.1
```
eyenx, line 5: Thanks for your contribution!

As this is an additional feature, please bump to 0.10.0

eyenx (changes requested):
Thanks for the contribution!

Please do a higher Chart version bump and please run pre-commit for generating the docs.

eyenx:
For the pre-commit hook pls see https://github.com/adfinis/helm-charts/blob/main/docs/development.md

charts/barman/Chart.yaml
eyenx: We are also missing the `artifacthub.io/changes` annotations in the Chart.yaml

See https://github.com/adfinis/helm-charts/pull/901 for a good example.

Thank you!

charts/barman/values.yaml
```diff
@@ -32,6 +32,20 @@ deployment:
   imagePullSecrets: []
   # -- Specify additional ENVs for the Deployment
   additionalENVs: {}
+  # Init containers to add to deployment's pod spec. At least one plugin provider image is required.
+  # If the value is a string then it is evaluated as a template.
```
eyenx, line 36:
```suggestion
  # If the value is a string then it is evaluated as a template.
  # -- Specify initContainers for the deployment
```

eyenx: We need this to generate the README.md you modified manually.

kfirfer: New to helm docs
done (I guess)

charts/barman/templates/deployment.yaml
```diff
@@ -81,6 +81,14 @@ spec:
           {{- end  }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+{{- if .Values.deployment.initContainers }}
+      initContainers:
+        {{- if eq (typeOf .Values.deployment.initContainers) "string" }}
```
eyenx, line 86: Can you quickly explain why we would be needing using a template for setting the initContainers? Is it not enough just rendering the array toYaml?

I see this construct also in the official velero Chart.

kfirfer: I just copied pasted from Velero chart
Don't know the reason for that
Should I use only array ?

kfirfer: Hmm I guess I know the reason
if you're templating over templating(e.g. helmfile gotmpl files), and you already has a string based value in other charts

eyenx: Makes sense

eyenx (approved):
LGTM thank you for your contribution!

eyenx:
Release is here

https://github.com/adfinis/helm-charts/releases/tag/barman-0.10.0

PR #958, opened 2023-02-09
feat(timed): add buildProject value for backend

trowik:
# Description

add buildProject value for backend

# Issues

# Checklist

### DIFF
--- a/charts/timed/values.yaml
+++ b/charts/timed/values.yaml
@@ -104,6 +104,8 @@ backend:
     admins: []
     # -- Define allowed domains for CORS
     corsAllowedOrigins: []
+    # -- Define name for build projects
+    buildProject: " BUILD"
     hurricane:
       # -- request queue length before hurricane denies requests and toggles readiness
       reqQueueLen: 250
--- a/charts/timed/Chart.yaml
+++ b/charts/timed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.15.3
+version: 0.15.4
 keywords:
   - timed
   - tracking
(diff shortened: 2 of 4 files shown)


trowik:
not sure about the chart version

charts/timed/values.yaml
```diff
@@ -104,6 +104,8 @@ backend:
     admins: []
     # -- Define allowed domains for CORS
     corsAllowedOrigins: []
+    # -- Define name for build projects
+    buildProject: " BUILD"
```
hairmare, line 108: is this whitespace legit?
```suggestion
    buildProject: "BUILD"
```
what's a build project?

hairmare:
you might also need to add something in `charts/timed/templates/configmap-backend.yaml` for this flag to do it's thing

charts/timed/values.yaml
trowik: The space there is on purpose.

Not really sure, but it's part of the Plan-Build-Run Model.
As far as I know we define them by using different cost centers in Timed , e.g. "DEVELOPMENT BUILD"

hairmare:
> not sure about the chart version

it's semver, so 0.x means you can do whatever you want. we tend to mostly increment the minor and only really update the patch release for actual fixes. at some point when charts start reaching 1.x, we'll have some real rules...

https://semver.org/

trowik:
> you might also need to add something in `charts/timed/templates/configmap-backend.yaml` for this flag to do it's thing

The value is just an env var that I already added to `config-backend.yaml`

trowik:
> > not sure about the chart version
> 
> it's semver, so 0.x means you can do whatever you want. we tend to mostly increment the minor and only really update the patch release for actual fixes. at some point when charts start reaching 1.x, we'll have some real rules...
> 
> https://semver.org/

alright, so 0.16.0 then

charts/timed/values.yaml
hairmare: oh, so the commercial "BUILD" and not something technical :sweat_smile: 

Do you know if maybe the existing `.redmine.buildProject` should be merged into this variable? 

https://github.com/adfinis/helm-charts/blob/dd9d46a7907c92f46615e018bdeee1f562410de0/charts/timed/templates/configmap-backend.yaml#L77

trowik: that one defines the name in Redmine, that is different from the one used in Timed

hairmare (approved):

trowik:
forgot to change the chart version...

hairmare:
> forgot to change the chart version...

doesn't matter that much, that's the nice thing about 0.x :)

PR #1002, opened 2023-03-23
feat: init keycloak-operator chart

hairmare:
# Description

After upstream declined accepting the offer of a community maintained keycloak-operator Helm chart I took a look at existing charts and came to the conclusion that most of them either support the legacy operator or are far from complete because they are either outdated or don't contain a clean way to manage an instance of the Keycloak CRDs as part of the initial deployment. This closes that gap by creating, imo fully featured, chart from scratch. In additional it also allows managing the realmimport CRD. AFAICT this will allow us to create ephemeral Keycloak installation with ie. identitiy federation. I still plan on doing some more tests with this deployment so we are sure that it really works, but the contents of this PR already work with chart-testing so it seems that they are already pretty close to being a working solution.

# Issues

# Checklist

* [x] ~add some simple pods for `helm test` that call the operator service and/or the included keycloak~ (there is too much operator in the loop for this to work)
* [x] figure out why the operator needs a service and potentially remove it or make it optional o_O


tongpu:
I won't find time to look at this before my holidays.

tongpu (commented):
One quick comment. Didn't yet find time for more.

charts/keycloak-operator/Chart.yaml
```diff
@@ -0,0 +1,18 @@
...
+description: Deploy Keycloak Operator and Keycloak
+type: application
+version: 0.1.0
+appVersion: "21.0.2"
+home: https://www.keycloak.org
+sources:
+  - https://github.com/keycloak/keycloak-k8s-resources
+  - https://github.com/adfinis-sygroup/helm-charts/tree/main/charts/keycloak-operator
+maintainers:
+  - name: adfinis-sygroup
```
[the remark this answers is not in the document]
hairmare: it's indicative of how long it took me to circle back to this chart and i'm fixing it in 077512c92d406b2411cb063dc127b8a6d756d33b.

eyenx:
Will look at it tomorrow

charts/keycloak-operator/templates/operator/clusterrole.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - keycloaks/status
+      - keycloaks/finalizers
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - delete
+      - patch
+      - update
```
tongpu, line 21: I think giving the operator all permissions would be acceptable here:

```suggestion
      - '*'
```

charts/keycloak-operator/templates/operator/clusterrole.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - keycloakrealmimports/status
+      - keycloakrealmimports/finalizers
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - delete
+      - patch
+      - update
```
tongpu, line 43: I think giving the operator all permissions would be acceptable here:

```suggestion
      - '*'
```

charts/keycloak-operator/templates/operator/deployment.yaml
```diff
@@ -0,0 +1,78 @@
...
+              path: /q/health/ready
+              port: http
+              scheme: HTTP
+            initialDelaySeconds: 0
+            periodSeconds: 30
+            successThreshold: 1
+            timeoutSeconds: 10
+          resources:
+            {{- toYaml .Values.operator.resources | nindent 12 }}
+      {{- with .Values.operator.nodeSelector }}
```
tongpu, line 67: If multiple instances of the operator can be deployed it might be a good idea to also add support for [podTopologySpreadConstraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/)

charts/keycloak-operator/values.yaml
```diff
@@ -0,0 +1,283 @@
...
+  # -- Configuration of the Keycloak server expressed as a keys and values that can be either direct values or references to secrets.
+  additionalOptions: ~
+  # - name: string
+  #   value: string
+  #   secret:
+  #     optional: bol
+  #     key: string
+  #     name: string
+
+  db:
```
tongpu, line 157: What's your opinion regarding adding a database chart as a dependency? It would make local development easier, but on the other hand some people might use in in production without knowing what they're getting themselves into.

charts/keycloak-operator/values.yaml
```diff
@@ -0,0 +1,283 @@
...
+    # -- The maximum size of the connection pool.
+    poolMaxSize: ~
+    # -- The database vendor.
+    vendor: ~
+    # -- Sets the database name of the default JDBC URL of the chosen vendor. If the `url` option is set, this option is ignored.
+    database: ~
+    # -- The minimal size of the connection pool.
+    poolMinSize: ~
+  features:
+    # -- Disabled Keycloak features
```
tongpu, line 187: It might be a good idea to add a link to the upstream docs here? Or is everything related to Keycloak configuration expected to be known by everyone deploying this service?

charts/keycloak-operator/templates/operator/service.yaml
tongpu, line 1: Is the service for the operator even required? Why would we talk to the operator via HTTP?

tongpu (commented):
Thanks a lot for your work! I think it looks pretty good already. Mostly some nitpicking comments.

charts/keycloak-operator/templates/operator/clusterrole.yaml
hairmare: It's based on the upstream clusterrole from [here](https://github.com/keycloak/keycloak-k8s-resources/blob/02a2913e2ad9e3454844bd8f6bb0ebfb03765c0e/kubernetes/kubernetes.yml#L137-L155), so i'd rather stick to that.

charts/keycloak-operator/templates/operator/clusterrole.yaml
hairmare: same here, it's based on [this](https://github.com/keycloak/keycloak-k8s-resources/blob/02a2913e2ad9e3454844bd8f6bb0ebfb03765c0e/kubernetes/kubernetes.yml#L157-L175)

charts/keycloak-operator/templates/operator/service.yaml
hairmare: It's most likely used for monitoring. i was never really able to confirm this, hence i made it optional via `operator.service.enabled` so we can figure out if it's needed later.

charts/keycloak-operator/values.yaml
hairmare: I've been thinking about that as well, but decided to not include a DB chart in the 0.1.0 release

charts/keycloak-operator/values.yaml
hairmare: I added a link in 96b6799a8537d05766a343bc2f68667465647a8d but i also assume some familiarity with keycloak and it's operator. I'm assuming this will hold true until the operator is more mature and reaches the auto pilot capability level.

tongpu (approved):
Looks good to me! Thanks a lot for your effort.

PR #1028, opened 2023-04-27
feat(misc-apps/reloader): new chart addition

eyenx:
# Description

Reloader helps us perform automatically a Rolling upgrade on the relevant DeploymentConfig, Deployment, Daemonset, Statefulset and Rollout if a ConfigMap or Secret used by it is changed. 

This would help us do rolling upgrades for some apps which need to be restarted upon secret change (Example Certificate is managed by cert-manager etc.)

# Issues
n/a

# Checklist

### DIFF
--- a/charts/misc-apps/Chart.yaml
+++ b/charts/misc-apps/Chart.yaml
@@ -20,5 +20,5 @@ annotations:
       description: |
         reloader: addition to misc-apps
       links:
-        - name: GitHub PR
-          url: https://github.com/adfinis/helm-charts/pull/1028
+        - name: GitHub project
+          url: https://github.com/stakater/Reloader
--- a/charts/misc-apps/Chart.yaml
+++ b/charts/misc-apps/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: misc-apps
 description: Argo CD app-of-apps config for miscellaneous small tools
 type: application
-version: 0.32.0
+version: 0.33.0
 home: https://github.com/adfinis/helm-charts/tree/main/charts/misc-apps
 sources:
   - https://github.com/adfinis/helm-charts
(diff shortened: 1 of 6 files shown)


charts/misc-apps/Chart.yaml
```diff
@@ -17,14 +17,8 @@ dependencies:
...
-          url: https://github.com/kedacore/keda/releases/tag/v2.10.0
-    - kind: changed
-      description: "keda: update chart from 2.9.3 to 2.10.1"
-      links:
-        - name: "feat: Ship KEDA Core v2.10.1 and align Kubernetes expectations"
-          url: https://github.com/kedacore/charts/pull/407
-        - name: "feat: prepare v2.10"
-          url: https://github.com/kedacore/charts/pull/400
+        - name: GitHub PR
+          url: https://github.com/adfinis/helm-charts/pull/xxx
```
hairmare, line 24: no need for self referential linking, please add some links that help discover reloader instead

charts/misc-apps/examples/reloader.yaml
```diff
@@ -0,0 +1,10 @@
+# reloader watching namespaces labeled with reloader=true and exporting metrics
+
+reloader:
+  enabled: true
+  project: infra-reloader
+  values:
+    reloader:
+      namespaceSelector: "reloader=true"
```
hairmare, line 8: is this really a string? why doesn't it use k8s api style specs?

hairmare:
also, can you add a description in this issue that explains what reloader is etc.

hairmare:
oh, and this seems to be missing an update to values.yaml and README.md as well as ci/default-values.yaml.

eyenx:
> oh, and this seems to be missing an update to values.yaml and README.md as well as ci/default-values.yaml.

YES cause I'm still working on this. Put in draft now. Sorry for the noise.

charts/misc-apps/examples/reloader.yaml
eyenx: 6213f5c

charts/misc-apps/Chart.yaml
eyenx: 4b2ce88

charts/misc-apps/examples/reloader.yaml
hairmare: is reloader just a kubectl wrapper? wouldn't a real k8s api client specify it as follows:
```yaml
namespaceSelector:
  reloader: "true"
```

eyenx: You are right. But the client expects this:

https://github.com/stakater/Reloader/blob/master/README.md?plain=1#L217

and the value is just passed over as argument in the deployment so...

gianklug (approved):
LGTM

PR #1061, opened 2023-06-28
infra-apps: actually update mimir-distributed

tongpu:
# Description

I forgot to actually update mimir-distributed in the previous PR #1060. This is the follow-up to bump

# Issues

Check if your setup requires [zone aware replication[(https://grafana.com/docs/mimir/latest/configure/configure-zone-aware-replication/) and disable it if not required (`ingester.zoneAwareReplication.enabled=false`, `store_gateway.zoneAwareReplication.enabled=false`).

Additionally verify that your Mimir deployment used structuredConfig instead of config, otherwise the upgrade might fail.

# Checklist

### DIFF
--- a/charts/infra-apps/Chart.yaml
+++ b/charts/infra-apps/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: infra-apps
 description: Argo CD app-of-apps config for infrastructure components
 type: application
-version: 0.164.0
+version: 0.165.0
 home: https://github.com/adfinis/helm-charts/tree/main/charts/infra-apps
 sources:
   - https://github.com/adfinis/helm-charts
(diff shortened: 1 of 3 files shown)


eyenx (approved):
bad reviewer I would say

charts/infra-apps/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: infra-apps
 description: Argo CD app-of-apps config for infrastructure components
 type: application
-version: 0.164.0
+version: 0.165.0
```
eyenx, line 5: shouldn't we do patchfix?

tongpu: I decided against it, as a new minor release better communicates that it's a new version.

eyenx:
LGTM

PR #1063, opened 2023-07-03
feat(prometheus-pushgateway)

luk43:
# Description
* introducing the prometheus pushgateway.
* Sadly it is not part of the current KPS stack
  * https://github.com/prometheus-community/helm-charts/issues/2030

I would need to add a config example. I need to work this out.

# Issues

# Checklist

### DIFF
--- a/charts/infra-apps/templates/prometheus-pushgateway.yaml
+++ b/charts/infra-apps/templates/prometheus-pushgateway.yaml
@@ -1,0 +1,37 @@
+{{ if .Values.prometheus-pushgateway.enabled }}
+{{ template "argoconfig.application" (list . "infra-apps.prometheus-pushgateway") }}
+{{ end }}
+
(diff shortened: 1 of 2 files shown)


hairmare:
the docs say this:
> ... infra-apps chart which contains apps we consider valuable on most clusters ...

IMO this doesn't fit that bill

bobs-bits:
I suggest a use case is for Prometheus Exporter in the stack is providing an endpoint for batch jobs to report to (e.g., Kubernetes Jobs).  For instance, I would like like a job to report through to a dashboard the number of good and bad records to recieved, and to generate alerts when it receives bad records, and or the number of records drastically changes.

It looks like https://github.com/latchmihay got a start on a push gateway integration, perhaps this could be merged into the stack (by someone more skilled than I...apologies for not raising a PR myself)?

https://github.com/latchmihay/kube-prometheus-pushgateway

MeCode4Food:
Much like having single instance jobs is a common use case, having a destination for logs for batch jobs to run is also a very common use case. 

Considering that kubernetes has support for single run `Job`s, having pushgateway added to the stack is a good support for logs from such `Job`s

PR #1073, opened 2023-07-20
chore(thanos): make fail

eyenx:
# Description

Make Thanos templating Fail

According to `docs/argocd-app-of-apps.md`:

> We will still keep the app available in the app-of-apps chart for at least 6 months so users can take time to replace it. During this time the component does not get regular updates unless there are egregious security issues that need addressing.
> Once we are ready to fully remove it we replace the components template with a fail template:

The [PR 854](https://github.com/adfinis/helm-charts/pull/854) keeping the app with just a DEPRECATION notice is from 8 months ago. So we make it fail now and wait for a year. 

# Issues

# Checklist

### DIFF
--- a/docs/argocd-app-of-apps.md
+++ b/docs/argocd-app-of-apps.md
@@ -163,7 +163,7 @@ Once we are ready to fully remove it we replace the components template with a `
 
 ```yaml
 {{ if .Values.component.enabled }}
-{{ fail("component is DEPRECATED, use "improved solution" instead") }}
+{{ fail "component is DEPRECATED, use "improved solution" instead" }}
 {{ end }}
 ```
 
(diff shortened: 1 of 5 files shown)


docs/argocd-app-of-apps.md
```diff
@@ -163,7 +163,7 @@ Once we are ready to fully remove it we replace the components template with a `
 
 ```yaml
 {{ if .Values.component.enabled }}
-{{ fail("component is DEPRECATED, use "improved solution" instead") }}
+{{ fail "component is DEPRECATED, use "improved solution" instead" }}
```
tongpu, line 166:
```suggestion
{{ fail "component is DEPRECATED, use improved solution instead" }}
```
The double quotes around `improved solution` are going to cause issues at some point. :D

hairmare (approved):

tongpu (approved):

PR #1135, opened 2023-10-12
feat(misc-apps): Update chartmuseum from 3.9.2 to 3.10.1

maxthier:
# Description
Update chartmuseum from 3.9.2 to 3.10.1

# Issues
https://github.com/adfinis/helm-charts/issues/1120

# Checklist


hairmare:
> I found the issue and therefore updated it but I don't know how I could test it. So it would be good if somebody can tell me how and what I can test or if somebody else could do that.

I'm not sure if we currently have an env that is well suited for testing this. So let's just merge it as is and then work on deprecating it next :scream: 

The future should be OCI published charts IMO

hairmare (approved):

Issue #1058, opened 2023-06-26
Exchange loki-stack with loki helm chart, due to deprecation of loki-stack


janaurka:
So, in `infra-logging` we are using [`loki-stack`](https://github.com/adfinis/helm-charts/blob/main/charts/logging-apps/values.yaml#L12-L14) as helm chart. This is [deprecated and should not be used anymore](https://community.grafana.com/t/difference-between-helm-charts-loki-and-loki-stack/87380). 

The new chart to use is [`loki`](https://github.com/grafana/loki/tree/main/production/helm/loki).

This only deploys `loki` and not `promtail` and `grafana`, so we should also add the `promtail` helm chart (and maybe [`vector`](https://vector.dev/docs/reference/configuration/sinks/loki/) -> the thing used by RedHat for OpenShift to ship logs to loki) to re-add the functionality loss we have from going `loki-stack` -> `loki`. 

Also it seems we don’t have a `grafana` support atm. We also need that, since `loki-stack` gave us that, but i overall like it way more as a standalone in `infra-apps` anyways.

janaurka:
https://grafana.com/blog/2022/12/19/the-only-helm-chart-you-need-for-grafana-loki-is-here/ more info

eyenx:
Integrating loki into grafana can be achieved over kubePrometheusStack and grafana.additionalDatasources.

tongpu:
I think the bare minimum functionality we should provide is combining the loki chart with the promtail chart. This should be rather easy, because the defaults of the promtail chart are very well suited to work with the loki chart and by combining them in the infra-logging chart as the two default options we should be well suited to replace our current deployment based on loki-stack.

Regarding Vector, I had a discussion with @vmaillot to replace a beats deployment at a customer with vector, so adding it to infra-logging is probably a good idea, but we need to evaluate if it is a good idea to make it the default over promtail and I would base that decision partially on resource consumption. It drives me crazy to see a simple component like promtail consume 100-300 MiB of memory and if Vector would be significantly better, we could discuss if it is better suited as an alternative. I expect that Vector would require a lot more configuration then a typical promtail setup, so I wouldn't consider it our first choice.

hairmare:
so, my understanding is that it's probably best to 
1. add loki and promtail to logging-apps
2. mark loki-stack as deprecated following our deprecation policy for -apps

[...]

PR #1146, opened 2023-11-14
feat(infra-logging): add extraObjects field

norbertgruszka:
# Description

Latest version of  `loki` helm chart comes with `extraObjects` variable which can be used to deploy additional resources. However, `loki-stack` chart must stayed behind, because mentioned variable is not present. Because of that I added `extraObjects` field to `infra-logging` chart.

# Issues

Unable to deploy additional resources alongside `loki-stack`.

# Checklist

### DIFF
--- a/charts/logging-apps/Chart.yaml
+++ b/charts/logging-apps/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: logging-apps
 description: Argo CD app-of-apps config for logging applications
 type: application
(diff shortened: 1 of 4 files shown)


hairmare:
IMO we should do https://github.com/adfinis/helm-charts/issues/1058 instead of this

tongpu:
I think before monkeypatching features into `logging-apps` we should progress with finally replacing `loki-stack` with `loki` according to #1058.

PR #1157, opened 2023-12-04
Fix VaultAutopilotHealthy alert name and time window in alert descriptions

in0rdr:
I would like to improve the alert name for the autopilot.

* https://developer.hashicorp.com/vault/docs/internals/telemetry/metrics/all#vault-autopilot-healthy
* https://developer.hashicorp.com/vault/docs/internals/telemetry/metrics/all#vault-autopilot-node-healthy

Also, I noticed a link that no longer points to a meaningful location. We can remove it.

### DIFF
--- a/charts/vault-monitoring/values.yaml
+++ b/charts/vault-monitoring/values.yaml
@@ -54,9 +54,7 @@ prometheusRules:
   # -- set of prometheus alerts to define
   # @default -- list of predefined alerts
   rules:
-    - alert: VaultAutopilotNodeHealthy
-      # Set to 1 if Autopilot considers all nodes healthy
-      # https://www.vaultproject.io/docs/internals/telemetry#integrated-storage-raft-autopilot
+    - alert: VaultAutopilotHealthy
       expr: vault_autopilot_healthy < 1
       for: 1m
       labels:
--- a/charts/vault-monitoring/values.yaml
+++ b/charts/vault-monitoring/values.yaml
@@ -93,7 +93,7 @@ prometheusRules:
         severity: critical
       annotations:
         summary: High frequency of failed Vault requests
-        description: There has been an increased number of failed Vault requests in the last 15 minutes
+        description: There has been an increased number of failed Vault requests in the last 20 minutes
     - alert: VaultResponseFailures
       expr: increase(vault_audit_log_response_failure[5m]) > 0
       for: 15m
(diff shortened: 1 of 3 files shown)


pree (approved):
Thanks, please also update `security-apps` to include the new version.

pree:
@adfinis/helm-charts what do you think about bumping this chart to `1.0.0` ?

eyenx:
> @adfinis/helm-charts what do you think about bumping this chart to `1.0.0` ?

Hmm read fix and this big bump and was like why? Are you sure this stuff is 1.0 ready?

in0rdr:
Looking at https://semver.org/#spec-item-5, we only need to care about the "breaking" change if we have `1.*`? Because probably nobody takes your word for it before `1.*` anyways (i.e., it's not yet a public API so you can do whatever). Let me revert to `0.3.0`, I was not aware of that rule nr. 5

charts/vault-monitoring/values.yaml
```diff
@@ -91,28 +89,28 @@ prometheusRules:
...
+        description: There has been an increased number of failed Vault requests in the last 20 minutes
     - alert: VaultResponseFailures
       expr: increase(vault_audit_log_response_failure[5m]) > 0
       for: 15m
       labels:
         severity: critical
       annotations:
         summary: High frequency of failed Vault responses
-        description: There has been an increased number of failed Vault responses in the last 15 minutes
+        description: There has been an increased number of failed Vault responses in the last 20 minutes
```
pree, line 100: How has this changed? It's still `for: 15m`

pree (changes requested):

tongpu:
> Looking at https://semver.org/#spec-item-5, we only need to care about the "breaking" change if we have `1.*`? Because probably nobody takes your word for it before `1.*` anyways (i.e., it's not yet a public API so you can do whatever). Let me revert to `0.3.0`, I was not aware of that rule nr. 5

I honestly believe that we depend on too many downstream charts in our `*-apps` charts that it's going to be close to impossible to ever provide a stable public API.

hairmare:
> I honestly believe that we depend on too many downstream charts in our `*-apps` charts that it's going to be close to impossible to ever provide a stable public API.

i believe that as well, but this is vault-monitoring hence the reasoning doesn't apply

PR #1196, opened 2024-02-06
feat(vault-monitoring): make bearerTokenFile optional

pree:
# Description

When using an unauthenticated access service for monitoring we don't need the bearerTokenFile.

# Issues

# Checklist

### DIFF
--- a/charts/vault-monitoring/templates/servicemonitor.yaml
+++ b/charts/vault-monitoring/templates/servicemonitor.yaml
@@ -24,8 +24,8 @@ spec:
       port: {{ .Values.vault.portName }}
       scheme: {{ .Values.vault.scheme }}
       scrapeTimeout: {{ .Values.vault.serviceMonitor.scrapeTimeout }}
-      {{- if .Values.vault.serviceMonitor.bearerTokenFile }}
-      bearerTokenFile: {{ .Values.vault.serviceMonitor.bearerTokenFile }}
+      {{- with .Values.vault.serviceMonitor.bearerTokenFile }}
+      bearerTokenFile: {{ . | quote }}
       {{- end }}
       tlsConfig:
         {{- if .Values.vault.ca }}
--- a/charts/vault-monitoring/templates/servicemonitor.yaml
+++ b/charts/vault-monitoring/templates/servicemonitor.yaml
@@ -24,9 +24,9 @@ spec:
       port: {{ .Values.vault.portName }}
       scheme: {{ .Values.vault.scheme }}
       scrapeTimeout: {{ .Values.vault.serviceMonitor.scrapeTimeout }}
-{{- if .Values.vault.serviceMonitor.bearerToken.enabled }}
-      bearerTokenFile: {{ .Values.vault.serviceMonitor.bearerToken.path }}
-{{- end }}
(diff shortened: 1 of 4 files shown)


charts/vault-monitoring/templates/servicemonitor.yaml
```diff
@@ -24,7 +24,9 @@ spec:
       port: {{ .Values.vault.portName }}
       scheme: {{ .Values.vault.scheme }}
       scrapeTimeout: {{ .Values.vault.serviceMonitor.scrapeTimeout }}
-      bearerTokenFile: '/etc/prometheus/config_out/.vault-token'
+{{- if .Values.vault.serviceMonitor.bearerToken.enabled }}
```
tongpu, line 27:
```suggestion
      {{- if .Values.vault.serviceMonitor.bearerToken.enabled }}
```

charts/vault-monitoring/templates/servicemonitor.yaml
```diff
@@ -24,7 +24,9 @@ spec:
...
+      bearerTokenFile: {{ .Values.vault.serviceMonitor.bearerToken.path }}
+{{- end }}
```
tongpu, line 29:
```suggestion
      {{- end }}
```

eyenx:
Can you annotate thechanges to the Chart.yaml

pree:
I added the changes, but unsure if this is needed as we can use the `serviceMonitor` from `vault-helm`. Externally hosted Vaults should never be contacted unauthenticated, so we might not want to merge this in.

charts/vault-monitoring/values.yaml
```diff
@@ -39,11 +39,7 @@ vault:
     labels: {}
     # -- scrapeTimeout of the serviceMonitor
     scrapeTimeout: 30s
-    bearerToken:
-      # -- enable bearerToken authentication
-      enabled: true
-      # -- path to the bearerToken
-      path: '/etc/prometheus/config_out/.vault-token'
+    # -- Specify a bearerTokenFile for authentication. Keep empty for Unauthenticated access
```
hairmare, line 42:
```suggestion
    # -- Specify a bearerTokenFile for authentication. Keep empty for Unauthenticated access
    bearerToken: ""
```

charts/vault-monitoring/values.yaml
```diff
@@ -40,6 +40,7 @@ vault:
     # -- scrapeTimeout of the serviceMonitor
     scrapeTimeout: 30s
     # -- Specify a bearerTokenFile for authentication. Keep empty for Unauthenticated access
+    bearerTokenFile: '/etc/prometheus/config_out/.vault-token'
```
hairmare, line 43: do you want the default to be "use a bearer token"? if so, then this is fine, if not, then an `""` might be preferential

hairmare (approved):

charts/vault-monitoring/values.yaml
pree: Yes, as it's highly recommended to secure the connection either by using the bearer token or securing the endpoint with other technologies (like network policies). Also, this way it's not a breaking change.

eyenx (approved):

PR #1214, opened 2024-04-12
chore(timed): update backend to v3.0.10

c0rydoras:
# Description

# Issues

# Checklist

### DIFF
--- a/charts/timed/Chart.yaml
+++ b/charts/timed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: timed
 description: Chart for Timed application
 type: application
-version: 0.16.32
+version: 0.16.33
 keywords:
   - timed
   - tracking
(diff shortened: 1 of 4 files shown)


hairmare (approved):

charts/timed/templates/service-backend.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   type: {{ .Values.backend.service.type }}
   ports:
     - port: {{ .Values.backend.service.externalPort }}
-      targetPort: http
+      targetPort: {{ .Values.backend.service.internalPort }}
```
hairmare, line 12: this might be a better fit for deployment-backend (so you can use name based ports from svc -> pod)

hairmare: so, it should work using `http` as before:

```suggestion
      targetPort: http
```

PR #1224, opened 2024-05-14
chore(timed): mark chart as deprecated

hairmare:
# Description

Chart was migrated to OCI registry and we can mark this as such here.

# Issues

# Checklist

### DIFF
--- a/charts/timed/Chart.yaml
+++ b/charts/timed/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: timed
-description: Chart for Timed application
+description: DEPRECATED use ghcr.io/adfinis/timed/helm/timed OCI Helm chart instead
 type: application
-version: 0.16.38
+version: 0.16.39
 keywords:
   - timed
   - tracking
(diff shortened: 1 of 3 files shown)


charts/timed/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: timed
-description: Chart for Timed application
+description: DEPRECATED use ghcr.io/adfinis/timed/helm/timed OCI Helm chart instead
 type: application
-version: 0.16.38
+version: 0.16.39
```
tongpu, line 5: I'd say deprecation should potentially be an increase of the minor version, but other then that it's fine.
```suggestion
version: 0.17.0
```

tongpu (approved):

charts/timed/Chart.yaml
hairmare: it's not a feature and it doesn't matter so imma be lazy and just merge it as is ;)

i do agree that other cases may warrant doing a patch release for deprecation, in this case the way forward is to switch to the 4.x stream of timed anyway (and sticking with this chart is in fact a bug).

PR #1233, opened 2024-05-22
Update service.yaml for bugfix

gulywwx:


hairmare:
It currently maps to the named port in the deployment:

https://github.com/adfinis/helm-charts/blob/6d0ba264d982e6d7553acf5f21f573dac8fb992d/charts/keycloak-operator/templates/operator/deployment.yaml#L42-L44

Did we miss something? Can you share the error you're encountering and what K8s version is affected?

gulywwx:
targetPort is supposed to be a port number in the Service resource. 

below is an example:

apiVersion: v1
kind: Service
metadata:
  name: my-service
spec:
  selector:
    app.kubernetes.io/name: MyApp
  ports:
    - protocol: TCP
      port: 80
      targetPort: 9376

hairmare:
A simple installation using `helm install kc adfinis/keycloak-operator -f charts/keycloak-operator/ci/operand-values.yaml` generates the following service:

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    meta.helm.sh/release-name: kc
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2024-05-22T11:26:00Z"
  labels:
    app.kubernetes.io/component: keycloak-operator
    app.kubernetes.io/instance: kc
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: keycloak-operator
    app.kubernetes.io/version: 23.0.4
    helm.sh/chart: keycloak-operator-1.0.3
  name: kc-keycloak-operator-operator
  namespace: default
  resourceVersion: "14983"
  uid: 94aa3dae-97d7-4a6a-a0aa-a74cfb476f4f
spec:
  clusterIP: 10.96.179.177
  clusterIPs:
  - 10.96.179.177
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: http
    port: 8080
    protocol: TCP
    targetPort: http
  selector:
    app.kubernetes.io/component: keycloak-operator
    app.kubernetes.io/instance: kc
    app.kubernetes.io/name: keycloak-operator
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```
Which maps to the port in the the matching deployment selected with `kubectl get pod -l 'app.kubernetes.io/component=keycloak-operator' -oyaml`:

```yaml
# ...
      ports:
      - containerPort: 8080
        name: http
        protocol: TCP
# ...
```

The Kubernetes docs contain an example of referencing ports by name in the [service port definition docs](https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports).

I'm not sure which K8s version introduced the possibility to reference ports by name. Is it possible that you are running this chart on a cluster that doesn't support referencing ports by name yet?

PR #1274, opened 2024-06-14
Include optional namespace in keycloak-operator templates

gwydionmv:
# Description
Included optional namespace field in the keycloak-operator templates.

# Checklist

* [X] This PR contains a description of the changes I'm making
* [X] I updated the version in Chart.yaml
* [X] I updated the changelog with an `artifacthub.io/changes` annotation in `Chart.yaml`, check the [example](docs/development.md#Changelog) in the documentation.
* [X] I updated applicable README.md files using  `pre-commit run`
* [X] I documented any high-level concepts I'm introducing in `docs/`
* [X] I am ready to test changes after they are applied and released


charts/keycloak-operator/templates/operator/clusterrole.yaml
```diff
@@ -3,6 +3,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: keycloakcontroller-cluster-role
+  {{- with .Values.operator.namespace }}
+  namespace: {{ . }}
+  {{- end }}
```
tongpu, line 8: ClusterRoles are not namespaced, so this is not required here.
```suggestion
```

charts/keycloak-operator/templates/operator/clusterrole.yaml
```diff
@@ -25,6 +28,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: keycloakrealmimportcontroller-cluster-role
+  {{- with .Values.operator.namespace }}
+  namespace: {{ . }}
+  {{- end }}
```
tongpu, line 33: ClusterRoles are not namespaced, so this is not required here.
```suggestion
```

charts/keycloak-operator/templates/operator/deployment.yaml
```diff
@@ -3,6 +3,9 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "keycloak-operator.operator.fullname" . }}
+  {{- with .Values.operator.namespace }}
+  namespace: {{ . }}
+  {{- end }}
```
tongpu, line 8: Is it also required that you can deploy the operator in another namespace the the one the Helm chart is being deployed into?
I can fully understand that you would want this for the keycloak instances, but the operator would be deployed in a central namespace right?

tongpu (changes requested):
I've left a few comments for you to consider.

Please make sure to also reflect the change in `values.yaml`, maybe with a commented out line showing how it can be used. Using `namespace: ""` might also work, because I think Helm treats empty string as if they don't exist.

PR #1317, opened 2024-09-12
Only create service with serviceMonitor

in0rdr:
Only create the service when `serviceMonitor.create` is set. The service is useless otherwise.

Background: I have a project where I can only use the `prometheusRules.enabled=true` but only want to use the prometheus rules, not the service monitor. Also, it creates a conflict in our deployment, because we already have a similar service with that name `vault-monitoring`.

### DIFF
--- a/charts/vault-monitoring/Chart.yaml
+++ b/charts/vault-monitoring/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: vault-monitoring
 description: monitor your vault server from within Kubernetes' prometheus
 type: application
-version: 0.4.2
+version: 0.5.0
 home: https://github.com/adfinis/helm-charts/tree/main/charts/vault-monitoring
 sources:
   - https://github.com/adfinis/helm-charts/tree/main/charts/vault-monitoring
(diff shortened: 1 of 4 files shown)


in0rdr:
I don't get what the check tries to tell me here:
```
Error: template: vault-monitoring/templates/service.yaml:1:14: executing "vault-monitoring/templates/service.yaml" at <.Values.serviceMonitor.create>: nil pointer evaluating interface {}.create
```

eyenx (approved):

eyenx (commented):
CI needs fixing

in0rdr:
I quickly tested these CI commands offline. It seems be working when I set `--target-branch` to my feature branch. The `ct lint` command then returns `chart version not ok. Needs a version bump!`, but at least no more nil pointer. Not sure how that's supposed to work though and I don't think messing with the [config file](https://github.com/adfinis/helm-charts/blob/main/hack/chart-testing/ct.yaml#L1) would be useful?

in0rdr:
I think there's a bug in that "Pluto" magic stuff.

```bash
$ tail -n1 ./hack/pluto.sh
helm template $chart_dir | pluto detect -
```
```bash
$ helm template ./charts/vault-monitoring
Error: template: vault-monitoring/templates/service.yaml:1:14: executing "vault-monitoring/templates/service.yaml" at <.Values.serviceMonitor.create>: nil pointer evaluating interface {}.create

Use --debug flag to render out invalid YAML
```

in0rdr:
@eyenx the `.Values` path was indeed wrong. CI is always right :stuck_out_tongue:

PR #1319, opened 2024-09-13
Check serviceMonitor.create for ConfigMap and Endpoint

in0rdr:
This checks if the serviceMonitor should be created before applying ConfigMap and Endpoint.

Example use case: Only install PrometheusRules


in0rdr:
Not yet finished, my pre-commit hook thingy is stuck again in some weird logic..

in0rdr:
Ready for review @4censord. I bumped revision to `0.6.0` because it's potentially breaking. This will remove the Endpoint and ConfigMap resources when `serviceMonitor.create=false`.

The default is `true`, so it's kind of save. Only peeps like me (with `serviceMonitor.create=false`) don't want to install any service monitor and just profit :moneybag: from the PromethesRules will not have unnecessary resources deployed anymore :muscle:

hairmare (approved):
LGTM

Issue #556, opened 2022-02-15
Deprecation of Gangway / finding a Replacement


eyenx:
Gangway is no longer mantained by VMWARE:

> VMware has ended active development of this project, this repository will no longer be updated.

We should find a replacement for it.

in #445 [dex-k8s-authenticator](https://github.com/mintel/dex-k8s-authenticator) was added to our security-apps chart.

This issue is for us to discuss a possible replacement by dex-k8s-authenticator or even finding a new candidate.

eyenx:
From my point of view:

The Dex <> Gangway way was mainly introduced in this charts due to SUSE CaaSP relying on it, as we new it was working well there.

I used dex-k8s-authenticator on a different customer's cluster. They are still using it now and it works well. It also works without adding additional SESSIONKEY's as secrets which makes it easier to deploy / mantain.

tongpu:
Probably we should approach this problem from two sides. For me as someone who has access to multiple clusters it would be beneficial to have a configuration that is compatible with [kubelogin/oidc-login](https://github.com/int128/kubelogin), because it would allow me to switch between clusters without having to open a web ui and integrate the shown configuration into my local kubeconfig.

But for other users (e.g. our customers) that only visit a single cluster a web UI might be beneficial. Dex would allow us to configure both OIDC clients at the same time allowing for both use cases.

hairmare:
> But for other users (e.g. our customers) that only visit a single cluster a web UI might be beneficial.

our customers rarely visit one single env, usually they have multiple envs like (dev, test, prod) as well.

hairmare:
maybe [this fork](https://github.com/tremolosecurity/kube-oidc-proxy) of [jetstack/kube-oidc-proxy](https://github.com/jetstack/kube-oidc-proxy) is something to keep an eye on

eyenx:
dex-k8s-authenticator is looking for new owners

https://github.com/mintel/dex-k8s-authenticator/issues/194

hairmare:
another option to look into is https://www.paralus.io/ (a CNCF sandbox project)

cfi2017:
I'd like to pick up this discussion again, following the deprecation of Gangway ( #1366 ). 

I like the idea of [kubelogin](https://github.com/int128/kubelogin), but it's a slightly different concept:
- We need higher level access to install it as we need to be able to change API server flags
[...]

PR #1366, opened 2025-01-16
chore(security-apps): Fail deprecated gangway installations

hairmare:
# Description

Gangway has been marked as deprecated for more than 6 months and now is the time to make it fail.

# Issues

# Checklist

### DIFF
--- a/charts/security-apps/Chart.yaml
+++ b/charts/security-apps/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: security-apps
 description: Argo CD app-of-apps config for security applications
 type: application
-version: 0.89.0
+version: 0.90.0
 home: https://github.com/adfinis/helm-charts/tree/main/charts/security-apps
 sources:
   - https://github.com/adfinis/helm-charts
@@ -15,10 +15,6 @@ dependencies:
     version: 0.9.1
     repository: https://charts.adfinis.com
 annotations:
-  artifacthub.io/containsSecurityUpdates: "true"
   artifacthub.io/changes: |
     - kind: changed
-      description: "chore: update oauth2-proxy from 6.10.1 to 7.8.2"
-      links:
-        - name: OAuth-Proxy 7.8.2 Release
-          url: https://github.com/oauth2-proxy/manifests/releases/tag/oauth2-proxy-7.8.2
+      description: "Fail installations of deprecated gangway component"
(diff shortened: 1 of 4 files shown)


charts/security-apps/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: security-apps
 description: Argo CD app-of-apps config for security applications
 type: application
-version: 0.89.0
```
cfi2017, line 5: Imo this should be a minor release, not a patch release

cfi2017: Else if someone is using 0.89.* their deployment may fail without user interaction

cfi2017 (changes requested):

charts/security-apps/Chart.yaml
hairmare: Are there still folks that do have `x.z.*`, i thought we deprecated that ages ago

cfi2017: It's a public chart, so we may not be the only consumers. It's also a breaking change in a patch version. A minor version is more of a signal that "hey, we may need to change something" for consumers so I believe it's a better fit here.

vmaillot:
https://git.adfinis.com/ch-cloudnative/team-coordination/-/issues/531

charts/security-apps/Chart.yaml
hairmare: Thanks for the review, i finally got to updating the PR to bump the minor and not the patch version.

I totally agree with you in principal, the main reason i asked was because we used to employ the `x.y.*` all over the place and we started retiring it's use a few years back.

charts/security-apps/templates/gangway.yaml
```diff
@@ -1,33 +1,3 @@
 {{ if .Values.gangway.enabled }}
-{{ template "argoconfig.application" (list . "security-apps.gangway") }}
+{{ fail "gangway is DEPRECATED, use dexK8sAuthenticator instead" }}
```
hairmare, line 2: **note:** I'm aware that the proposed replacement has been abandoned in the meantime.

The dex-k8s-authenticator recommendation is a verbatim copy of what has been in values.yaml for years.

I'm assuming we will replace the recommendation based on the results of #556.

cfi2017 (approved):
lgtm

PR #1521, opened 2025-10-03
fix(keycloak-operator): allow specifying false in ternary values

hairmare:
# Description

We were comparing false values using `with`, leading to settings like `startOptimized=false` getting removed instead of rendered.

This changes the check to just render them if available, even if the values are falsey.

# Issues

# Checklist


Maryana-B (approved):
I'm not sure why this is asking for my review. It looks good to me.

PR #1909, opened 2026-07-15
feat(backup-apps): Bumped back8sup chart to 0.5.0

### LINKED ISSUE #1813: New Version Available: back8sup 0.5.0
New version available for **back8sup**!

Current `targetRevision` set: `0.4.8`
New version available: `0.5.0`

Please consider creating a PR to update the `targetRevision` in this file: [charts/backup-apps/values.yaml](https://github.com/adfinis/helm-charts/blob/main/charts/backup-apps/values.yaml)

Remember to check the upstream chart for updated (or more stable) versions than the one mentioned while you address this ticket.

Thanks.

Xelef2000:
# Description
Bumped back8sup chart to 0.5.0, the last  version.
Also set `enabled: false` since this chart is deprecated.

# Issues
* fixes #1813

# Checklist

### DIFF
--- a/charts/backup-apps/Chart.yaml
+++ b/charts/backup-apps/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: backup-apps
 description: Argo CD app-of-apps config for backup components
 type: application
-version: 0.6.0
+version: 0.7.0
 home: https://github.com/adfinis/helm-charts/tree/main/charts/backup-apps
 sources:
   - https://github.com/adfinis/helm-charts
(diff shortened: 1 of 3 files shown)


Xelef2000:
@hairmare the chart was previously set to enabled in the values file (enabled: true), since we deprecated it, i set it to false. 
Why was it enable by default, and since velero is the official replacement, should we enable that by default?

hairmare:
> @hairmare the chart was previously set to enabled in the values file (enabled: true), since we deprecated it, i set it to false. Why was it enable by default, and since velero is the official replacement, should we enable that by default?

@Xelef2000 IMO `enabled: false` is the right thing to do. I was a bit surprised that it wasn't already set to `false`. I think all defaults should be `false`, to ensure that enabling and configuring a component is an active task and doesn't just happen magically. The backup-apps chart seems to have always had this issue.

hairmare (approved):

